PR #20, opened 2014-12-01
Servant pattern

sideris:
Adding the servant pattern and relevant diagrams(GenMyModel generated)


iluwatar:
@all-contributors please add @sideris for code

PR #31, opened 2015-03-14
Update App.java

mbirkenkamp:
it is important to note that the singleton pattern is inherently unsafe for distributed environments, since the initialization state is not mirrored across different instances of the java vm

### DIFF
--- a/singleton/src/main/java/com/iluwatar/App.java
+++ b/singleton/src/main/java/com/iluwatar/App.java
@@ -3,10 +3,16 @@ package com.iluwatar;
 /**
  *
  * Singleton pattern ensures that the class (IvoryTower) can have only one
- * existing instance per java vm instance and provides global access to that it.
- *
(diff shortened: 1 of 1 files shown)


iluwatar:
Thanks, this is a good point. Could you please add some more explanation about it? I see there is a good discussion about this here: http://stackoverflow.com/questions/17721263/singleton-across-jvm-or-application-instance-or-tomcat-instance

aceiro:
Hi all

Great job:)

Anyone have some opinion about evolve and/or write some design pattern for
Jdk 8, in particular, using new features existing on Jdk such as lambda
expression, around patterns and so on.

Let me know yours opinions.

Sorry if it's not a propriate discussion in this track.

Best regards
Erik

iluwatar:
Hi @aceiro,

Thanks for your comment. It's a good point, why not start using Java 8 right now? I suggest that you create an issue for the project. We can then see where Java 8 features could be used and what benefit does it bring.

PR #32, opened 2015-03-18
added Wizard pattern

vehpsr:


iluwatar:
Hey there, this stuff looks awesome!

wizard/src/main/java/com/iluwatar/App.java
```diff
@@ -0,0 +1,53 @@
...
+ * Object constructed by Wizard pattern address this issue with "fail fast" approach
+ * and trying to expose some initialization problems at compile time.
+ *
+ * Main disadvantage of Wizard pattern is verbosity and often complicated hierarchy of auxiliary classes and interfaces.
+ * Also intermediate return type may be not what is desired (in case you want to save it in some variable
+ * and pass into another function for further modification) or predictable as with Builder patter.
+ * But when done right (and in right places) can significantly simplify usage of some specific domain API.
+ *
+ * SqlBuilder wizard try to control flow (or just impose some restrictions) for query initialization process, like:
+ * - you can't call anything else than select() after initialization;
```
iluwatar: Are these restrictions implemented in the code?

wizard/src/main/java/com/iluwatar/App.java
```diff
@@ -0,0 +1,53 @@
...
+ * - final move, toSql(), can be called only if you reach certain point;
```
iluwatar: Are these restrictions implemented in the code?

wizard/src/main/java/com/iluwatar/App.java
```diff
@@ -0,0 +1,53 @@
...
+ * - if you call orderBy() then you can't call where() or from() afterwards, etc.
```
iluwatar: Are these restrictions implemented in the code?

iluwatar:
Is there some source (i.e. book, Wikipedia) that describes this pattern?

wizard/src/main/java/com/iluwatar/FromClause.java
```diff
@@ -0,0 +1,5 @@
+package com.iluwatar;
+
```
iluwatar: Please add short Javadoc description

wizard/src/main/java/com/iluwatar/QueryClosure.java
```diff
@@ -0,0 +1,7 @@
+package com.iluwatar;
+
```
iluwatar: Please add short Javadoc description

wizard/src/main/java/com/iluwatar/QueryFiltering.java
```diff
@@ -0,0 +1,6 @@
+package com.iluwatar;
+
```
iluwatar: Please add short Javadoc description

wizard/src/main/java/com/iluwatar/QueryOrder.java
```diff
@@ -0,0 +1,5 @@
+package com.iluwatar;
+
```
iluwatar: Please add short Javadoc description

wizard/src/main/java/com/iluwatar/QueryPagination.java
```diff
@@ -0,0 +1,6 @@
+package com.iluwatar;
+
```
iluwatar: Please add short Javadoc description

wizard/src/main/java/com/iluwatar/SelectClause.java
```diff
@@ -0,0 +1,5 @@
+package com.iluwatar;
+
```
iluwatar: Please add short Javadoc description

wizard/src/main/java/com/iluwatar/SqlBuilder.java
```diff
@@ -0,0 +1,161 @@
+package com.iluwatar;
+
+import java.util.AbstractMap;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map.Entry;
+
+import com.iluwatar.utils.CollectionUtils;
+
```
iluwatar: Please add short Javadoc description

wizard/src/main/java/com/iluwatar/WhereClause.java
```diff
@@ -0,0 +1,5 @@
+package com.iluwatar;
+
```
iluwatar: Please add short Javadoc description

wizard/src/main/java/com/iluwatar/utils/CollectionUtils.java
```diff
@@ -0,0 +1,20 @@
+package com.iluwatar.utils;
+
+import java.util.Collection;
+import java.util.Iterator;
+
```
iluwatar: Please add short Javadoc description

jesusvazquez:
I've been performing a few mins research about this. I haven't found anything related to that "Wizard" name for this kind of pattern. For me it looks like a custom Builder adapted to how SQL querys look like or how SQL syntax is made. Am i wrong?

vehpsr:
Hi guys.

So about origin of this pattern.
I can't direct you to some wiki entry or document or authority publication. I don't know any. If you Google "Wizard pattern" you mostly get answers that it is UI/UX approach of building interactive view dialogues.
Thou if you Google "Wizard pattern in java" you will find something: the same, reprinted in many blogs, virgin-married-hasChildren person-building example.
As far as I can say the origin is this article:
http://developer4life.blogspot.com/2012/01/wizard-design-pattern.html
I think so mostly because they have nice follow-up with using this pattern building RegEx Wizard:
http://developer4life.blogspot.com/2013/08/writing-java-regular-expression-without.html
and respectively code on GitHub at:
https://github.com/azarian/wizards

So is it pattern or just principle, does it really called "Wizard", and will any "train wrack" of chained commands will be example of one, I don't know.

¶

Nonetheless after second thought I decided to cancel this PullRequest.
The reason - I don't like implementation.
All this business with inner-impl classes is mostly redundant and don't contribute any real value (think this design is strongly influenced by original article).
The big picture as I see it now are more consistent and in-line with plain Builder pattern and contains of:
- class SqlQuery with private constructor that have single private field _sql.
- public inner class Builder that implements all necessary interfaces. Only Builder contains all auxiliary variables as _db, _limit, etc., that gets digested in build() method.

So main difference with plain Builder will be that Builder always return its type,
```java
    public Builder setValue(String val) {
        _val = val;
        return this;
    }
```
but Wizard will return some Builder subType.
```java
    public BuilderSubType setValue(String val) {
        _val = val;
        return this;
    }
```
The only problem that I see so far is Builder instantiation. I can't do this using constructor ```new SqlQuery.Builder()```, because it will return to generic composite type.
So the solution is to make Builder with private constructor as Singleton and static init() method with subType return type. And that solution I find not so "pure"... Need to play with it some time.

Also will pay more attention to javadocs.

Although I don't know how properly do amendments to PullRequest (this is mine first). Can I revert all changes and start from scratch or must commit changes on top of this implementation?

iluwatar:
> All this business with inner-impl classes is mostly redundant and don't contribute any real value (think this design is strongly influenced by original article).
The big picture as I see it now are more consistent and in-line with plain Builder pattern and contains of:

Yes, I agree with you. As such the pattern seems to very similar to Builder. With some more implementation (the business rules) the example would most likely develop into a combination of patterns Builder and State.

If you want to develop the code in this pull request to this direction then you should add it under Builder module as an additional example. Your choice if you want to do it in this pull request or another.

stephen-lazarionok:
Perhaps it would be useful. Looks like Apache Camel framework implements 'Wizard pattern' for building its routes:
```
RouteBuilder builder = new RouteBuilder() {
    public void configure() {
        errorHandler(deadLetterChannel("mock:error"));
 
        from("direct:a")
            .choice()
                .when(header("foo").isEqualTo("bar"))
                    .to("direct:b")
                .when(header("foo").isEqualTo("cheese"))
                    .to("direct:c")
                .otherwise()
                    .to("direct:d");
    }
};
```

iluwatar:
IMHO this is Builder pattern in work. 

See https://github.com/apache/camel/blob/0e195428ee04531be27a0b659005e3aa8d159d23/camel-core/src/main/java/org/apache/camel/builder/RouteBuilder.java and https://github.com/apache/camel/blob/01f08a6267232089adba90a380050ba5dc9c43d2/camel-core/src/main/java/org/apache/camel/builder/BuilderSupport.java.

Compare to https://github.com/iluwatar/java-design-patterns/blob/master/builder/src/main/java/com/iluwatar/App.java.

Great addition to "Real world examples" list of Builder though.

vehpsr:
As in case with State vs Strategy
> While the implementation is similar they solve different problems. The State pattern deals with what state an object is in - it encapsulates state-dependent behavior. The Strategy pattern deals with how an object performs a certain task - it encapsulates an algorithm.

Maybe intent can be also a difference.
- Builder can solve performance problem (canonic example, I think, is StringBuilder at work: collect references and then create object at once)
- Builder can have single validation point -- build() method (to achieve this goal you must use only constructor injection for POJO)
- Building immutable objects (of course which is often the case)
- Funky syntax :)
- ...

Thou I agree that in real life we most of the time have some blended version.

PR #33, opened 2015-03-21
Update App.java

fx773d:
changed List to ArrayList.


iluwatar:
Just one question. Why would you want to change the `List` to `ArrayList` in this case?

fx773d:
Well, using List will be useful for further implementations with different kind of Lists and that means more maintainable code. Nevertheless in that way, whenever the user wants to use an ArrayList specific method, he/she must explicit casting the variable and that might be cause exceptions if used without "instanceof" condition. Briefly, I think It is matter of taste, congratulations for your successful work.

Regards.

iluwatar:
@fx773d  I'm sorry but I need to reject this pull request because the change doesn't bring any benefits.

PR #37, opened 2015-03-27
Callback pattern description amendment

vehpsr:
Guess this description will be more close to accurate.

Also come up with kind of controversial **Real world example**: one of [CyclicBarrier] (http://docs.oracle.com/javase/7/docs/api/java/util/concurrent/CyclicBarrier.html#CyclicBarrier%28int,%20java.lang.Runnable%29) constructors can accept callback that will be triggered every time when barrier is tripped.
Though ideologically, I think, it is good Callback example its very differ in implementation.
Would it be appropriate to mention this case in description?


iluwatar:
Yes, I think CyclicBarrier example would be good to mention. I will merge this after you have done the change.

PR #39, opened 2015-03-29
added Property pattern

vehpsr:
Based on Steve's Yegge [The Universal Design Pattern] (http://steve-yegge.blogspot.com/2008/10/universal-design-pattern.html) article


iluwatar:
> Today I thought I'd talk about a neat design pattern that doesn't seem to get much love: the Properties Pattern. In its fullest form it's also sometimes called the Prototype Pattern.

Did you check if this duplicates [Prototype](https://github.com/iluwatar/java-design-patterns#prototype)?

iluwatar:
In this [entry](http://www.codeproject.com/Articles/10103/Refactoring-to-Adaptive-Object-Modeling-Property-P) I can see it is a combination of prototypical instance creation + property collection.

iluwatar:
Wow, [that](http://steve-yegge.blogspot.fi/2008/10/universal-design-pattern.html) was a long read!

iluwatar:
Good work and thanks for pointing out the article. I feel enlightened :)

vehpsr:
Yeah, Steve prefers to call his posts essays because of their length, but if I had to name three blogs that made greatest influence on me as a developer, then the list is definite:
* [Joel Spolsky] (http://www.joelonsoftware.com/)
* [Jeff Atwood] (http://blog.codinghorror.com/)
* Steve Yegge

iluwatar:
Great list. I would add [Martin Fowler](http://www.martinfowler.com/) and [Uncle Bob](http://blog.cleancoder.com/).

PR #46, opened 2015-04-15
add redo function into command pattern

zhwbqd:
1. add redo function into command pattern.
2. add .iml and .idea into .gitignore

### DIFF
--- a/command/src/main/java/com/iluwatar/Command.java
+++ b/command/src/main/java/com/iluwatar/Command.java
@@ -11,6 +11,8 @@ public abstract class Command {
 
 	public abstract void undo();
 
+	public abstract void redo();
+
 	@Override
 	public abstract String toString();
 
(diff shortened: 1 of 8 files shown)


command/src/main/java/com/iluwatar/Command.java
```diff
@@ -11,6 +11,8 @@
 
 	public abstract void undo();
 
+	public abstract void redo();
```
vehpsr: just thinking...
can we make undo()/redo() (as well as execute()) return boolean instead of void, so we can't shrink if no target provided, target already shrinked (small) or target is immune. 
as result command can't be queued/dequeued if execute() fails.

iluwatar:
This looks good to me. The queues make the example more realistic.

iluwatar:
Returning boolean from undo()/redo()/execute() is ok too, if you want to add that. Let me know when you're done with the changes and I will merge the PR.

iluwatar:
Resolved with commit 58da95b4676ff09cbed607934ab3722063128e16

PR #47, opened 2015-04-15
Added DAO implementation, modified readme, and added UML diagram

mgiovenco:
Contribution of the DAO implementation pattern.  Includes code demonstrating basic CRUD, read updates, and diagram made using Eclipse plugin.


iluwatar:
Thank you! Looks good.

Issue #68, opened 2015-06-23
Async Method Invocation pattern


iluwatar:
https://en.wikipedia.org/wiki/Asynchronous_method_invocation

Issue #104, opened 2015-07-20
Coverage analysis fails


iluwatar:
[ERROR] Failed to execute goal org.eluder.coveralls:coveralls-maven-plugin:3.1.0
:report (default-cli) on project java-design-patterns: I/O operation failed: No
source found for domainapp/dom/modules/simple/QSimpleObject.java -> [Help 1]

iluwatar:
[ERROR] Failed to execute goal org.eluder.coveralls:coveralls-maven-plugin:3.1.0
:report (default-cli) on project java-design-patterns: Build error: 32 -> [Help
1]

trautonen:
Coveralls plugin fails because it cannot handle equally named classes in a multi-module project. There are at least three modules containing class `com.iluwatar.App` and the plugin fails to select a correct one for a coverage report. This is known defect in the plugin, but really hard to solve. Version 3.2.0 of the plugin will improve the error message of such cases so it's easier to detect, but still no real solution for the issue.

To get the coveralls reports correctly, make sure there are no same fully qualified class names between the modules. For example add a package suffix for each module. I think most of the modules follow this naming pattern already.

iluwatar:
Thank you for looking into the problem. I will take your hint and make sure the classes have different fully qualified domain names. This is related to issue #108

PR #106, opened 2015-07-22
Async method invocation

trautonen:
Implementation of #68.

I don't have eclipse installed so couldn't create the UML class diagram.


iluwatar:
Thanks Tapio! I will look into this.

iluwatar:
The code is excellent. Exemplary work!

iluwatar:
I added the class diagram.

iluwatar:
Tapio, do have an idea why #104 occurs?

trautonen:
I'll try to look on that tonight.

Issue #109, opened 2015-07-23
Half Sync/Half Async pattern


npathai:
This is a _concurrency_ pattern. The pattern separates asynchronous I/O from the synchronous one. Main thread doesn't block on incoming client requests and long-running operations are offloaded to a dedicated synchronous layer. Processing results are delivered by the means of callbacks.
Some resources:
 - http://www.cs.wustl.edu/~schmidt/PDF/PLoP-95.pdf
 - http://zezutom.blogspot.in/2014/08/concurrency-patterns-half-sync-half.html
 - http://www.dre.vanderbilt.edu/~schmidt/cs282/PDFs/6-Concurrency-and-Synchronization-part10.pdf

iluwatar:
Great find, thanks for reporting this.

npathai:
I am working on it.

npathai:
Note to self - we also need to add reference to POSA book with this task in README

iluwatar:
Please close this issue, if you are satisfied.

npathai:
Once #204 is merged we can close this.

PR #177, opened 2015-07-29
Half sync half async

npathai:
Added implementation of #109 with class diagram. @iluwatar Please review it. If the implementation is good to go then I just need to update the README.


iluwatar:
@npathai This looks great! Here's some remarks I made:
- add the module to the parent pom.xml
- reference the issue in your commit messages e.g. #109
- add the pattern to `README.md` also

npathai:
@iluwatar It's done

npathai:
Added reference in README for this pattern.

PR #178, opened 2015-07-29
Changed improper description about double dispatch in FAQ section

npathai:
The FAQ section about the difference between double dispatch and Visitor states that C++ and Java **support** this. But actually they do not support it. Have corrected the mistake.


iluwatar:
Well spotted :+1:

Issue #181, opened 2015-07-30
Migrate travis build to container based infrastructure


npathai:
I was viewing build logs when I found out that we are running build on legacy infrastructure and got a notification for migration to container based infrastructure.

[Travis CI - Why migrate to container based infrastructure](http://docs.travis-ci.com/user/migrating-from-legacy/?utm_source=legacy-notice&utm_medium=banner&utm_campaign=legacy-upgrade)

One of the best advantages is faster build times

mafagafogigante:
Shouldn't this be replicable in my fork?

I get `This job is running on container-based infrastructure, which does not allow use of 'sudo', setuid and setguid executables.`

npathai:
@mafagafogigante I have already migrated my fork to container based infrastructure. But Travis is facing issues downloading dependencies for `naked-objects-intergration` and the build is failing. I need to figure out why is that so. FYI https://travis-ci.org/npathai/java-design-patterns

iluwatar:
@npathai I think this is the same build issue I worked on yesterday, see https://github.com/iluwatar/java-design-patterns/commit/185f296a5b03f16edb62d0023441e89ad4365501. Try to catch up your fork from the master and hopefully the problem is fixed.

npathai:
@iluwatar Thanks that solved the issue. It's done. Generating PR.

mafagafogigante:
I had the same problem of @npathai.

Thanks for fixing it.

PR #193, opened 2015-08-07
PR for #181 Migrate to container based infrastructure

npathai:
It is a small change in travis.yml which migrates the build to container based infrastructure. Solves #181


iluwatar:
Thank you for looking into this.

PR #200, opened 2015-08-11
Provided better navigation in README

npathai:
Presently some of the sections such as FAQ, How to contribute were not easily accessible from the top of README. I didn't even realize they existed until I scrolled till very bottom. I think this navigation scheme provides easy access to all major sections of README.


iluwatar:
This is a great improvement @npathai Well spotted!

Issue #205, opened 2015-08-13
Better Structure / More Info


noobxgockel:
Hello there,

so i recently got the "Head First Design Patterns" book from my boss and i found a pretty nice Structure to put patterns into it. Maybe we could adapt to something like this. I mean the overview you get from this collection of patterns is really nice and helpfull but i find if i wanted to use some pattern mentioned here i havent got enough information to fully comprehend and implement the patterns correctly so i go to other pages like wikipedia or sorts.

The Structure the book mentions:
- Name + Classification/Category
- Intent / Pattern definition
- Motivation (Describes concrete scenarios, their problems and how this pattern solves the problems)
- Applicability (Situations in which the pattern can be applied)
- Structure (Diagram)
- Participants (Describes the classes and their respective roles in the pattern)
- Collaborations (How the participants work together)
- Consequences (The Effects this pattern has, good and bad)
- Implementation/Sample Code (Techniques, thought process, ...)
- Known Uses
- Related Patterns

This resource covers quite a lot of the mentioned points, but not all of it and not in a very detailed manner. So maybe this issue is something we could work towards, maybe not in the near future, but when we find time for it. But i believe making this happen will move us one step closer to beeing the best resource for java design patterns C:

Hope you like it ^^

npathai:
@noobxgockel Yes that surely will help. Because one thing is to understand the patterns, it is way different to understand where and how to implement them. As they never are implemented as island, they actually are in subtle ways linked and used with each other. That's [Pattern Language](https://en.wikipedia.org/wiki/Pattern_language) a term coined by Christopher Alexander for building architecture.
There can be another analogy of ingredients and recipe. Ingredients - individual patterns and the Recipe is pattern language (actual real life architecture).

What GoF book and POSA series tries to achieve with all these sections is to give enough information so that a language can be formed by using vocabulary of these individual patterns. So yes we should strive to provide as much information with a pattern as possible at least with new patterns we implement. I hope @iluwatar agrees with this.

What we also need to promote and document is, thinking of design patterns as a form of communication, and not over engineering the design by unnecessarily shoving in various design patterns just because we can. Many starters do this. Just remembering the YAGNI principle.

Any comments? @noobxgockel @iluwatar

iluwatar:
This is good to discuss. I think this has all to do with our vision and what we want to achieve with this project.

[...]

Issue #184, opened 2015-07-31
Fluent Interface pattern


iluwatar:
http://www.martinfowler.com/bliki/FluentInterface.html

http://stackoverflow.com/questions/31733262/what-is-this-the-name-of-this-java-state-based-design-pattern

hannespernpeintner:
I try to implement it :)

hannespernpeintner:
Sent a pull request https://github.com/iluwatar/java-design-patterns/pull/207 :)

npathai:
@hannespernpeintner Thanks for your PR. :+1: 
I have not read the whole example but just took a brief idea. @iluwatar Would like to hear your suggestion on "Should we implement this example as a `FluentIterable`". This will be a minimal mock of [Guava's FluentIterable](http://docs.guava-libraries.googlecode.com/git/javadoc/com/google/common/collect/FluentIterable.html). I think it makes the pattern much easily understandable. 
It is just my opinion that using that example would be very close to how it will be used in real applications.
A mockup
```java
FluentIterable.from(integerList)
.filter(negatives())
.first(10)
.max();
```

@hannespernpeintner There are many more real applications one of them is Guava, Java 8 Stream API are some of them which we should document.

npathai:
@iluwatar Or even better we can just use Java 8 Stream API to demonstrate it.

hannespernpeintner:
You are welcome :) 
@npathai guavas fluentiterable is a complete and concise Implementation and could be a receipt by itself, i dont think one should reimplement it. However, maybe its source should be linked in the read.me?
Using the stream API would be easy and possible but isnt the key of this project to show how the pattern can be Used - and *implemented* as well?

npathai:
@hannespernpeintner I agree with you. I think we should implement it by hand, a very minimal example. We are in the same dilemma as we faced with #93 . Whether we should implement it or just use existing library. There we have decided to use Camel for just showcasing the use of design pattern (because it would have required a lot of boilerplate framework classes to implement them). I was kinda in the inclined towards implementing them by hand. But after the discussion in #93 we decided to go against it.
Do you think we can come up with some better example?

npathai:
@iluwatar Do you think we should have a discussion round about the example that we will be demonstrating prior to implementation. Reason is that I find it useful if the example demonstrated is nearer to a real application. But that can be just me.
Again it can mean that our tasks run longer because it can be difficult to come to a consensus. I don't know just thinking out loud. Guys?

[...]

Issue #55, opened 2015-05-27
Repository


fernandojunior:
Great job.
How about repository? 
Reference: http://martinfowler.com/eaaCatalog/repository.html

iluwatar:
Thanks for pointing that out. It is definitely on the roadmap.

iluwatar:
@fernandojunior The example is now available.

fernandojunior:
I didn't know Spring has this functionality...
Thanks!

Issue #45, opened 2015-04-11
DAO Design Pattern


ozlerhakan:
Hi @iluwatar ,

First of all, thank you for this comprehensive repo!  I haven't seen the dao design pattern among the rest. So, I am just wondering whether it is actually part of the design solutions or not.

iluwatar:
Hi there @ozlerhakan !

I'm glad you find the repository interesting. The [DAO](http://en.wikipedia.org/wiki/Data_access_object) pattern certainly belongs here, we just don't have an implementation for it yet.

ozlerhakan:
Ah okay then, so If the current contributors (including you) will not plan to implement it, I can give you a hand for the implementation via PR but not now (:

iluwatar:
Contributions are welcome @ozlerhakan . Please leave the issue open and it will be fixed soon.

mgiovenco:
I believe I have the implementation for this ready to go.  If it's okay for me to contribute, I can create a pull request for this.

iluwatar:
@mgiovenco That is great! Please proceed with the pull request and we'll check it.

mgiovenco:
Will do!  I just need to add the diagram piece before creating pull request.  Expect the pull request in the next day or two.

iluwatar:
Resolved with commit 58da95b4676ff09cbed607934ab3722063128e16

ozlerhakan:
Thanks @iluwatar  and @mgiovenco :+1:

pavansn:
Hi @iluwatar ,
Can I add these design patterns into the repository?
          1. 2-way adapter
          2. IOC - Inversion of Control
          3. Dependency Injection
I hope these patterns belong to this repository.

iluwatar:
Hi @pavansn ,

Surely these patterns all belong here. You are welcome to contribute. I suggest you make an individual pull request out of each of the patterns so we can review and comment easily.

PR #206, opened 2015-08-14
Active record pattern implementation

stephen-lazarionok:


stephen-lazarionok:
Pull request #206 is prepared. Please review.

npathai:
I will review it this weekend.

active-record/pom.xml
```diff
@@ -0,0 +1,30 @@
...
+        <groupId>com.iluwatar</groupId>
+        <version>1.5.0</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>active-record</artifactId>
+
+    <dependencies>
+
+        <dependency>
```
npathai: Any particular reason you have chosen H2 and not Derby? Derby is packaged with Java so will not require external dependency.

active-record/src/main/java/com/iluwatar/active/record/MagicWand.java
```diff
@@ -0,0 +1,264 @@
...
+
+        return calculateWoodMagicFactor() * calculateWoodMagicFactor() * this.getLengthInches();
+    }
+
+    public void castFireball() throws SpellCastException {
+
+        if (this.calculateMagicPower() >= 10.0) {
+            System.out.println(this.toString() + ": a fireball spell is casted");
+        } else {
+            throw new SpellCastException("Can not cast lighting! Not enough magic power...");
```
npathai: A typo here, the message should be `Can not cast fireball!`.

active-record/src/main/java/com/iluwatar/active/record/MagicWand.java
```diff
@@ -0,0 +1,264 @@
...
+            case UNICORN_HAIR:
+                return 0.8d;
+            default:
+                return 0.0d;
+        }
+    }
+
+    public double calculateMagicPower() {
+
+        return calculateWoodMagicFactor() * calculateWoodMagicFactor() * this.getLengthInches();
```
npathai: You should be consistent with either using `this` for instance methods or not at all. I prefer not using them if not required.

active-record/src/main/java/com/iluwatar/active/record/MagicWand.java
```diff
@@ -0,0 +1,264 @@
...
+        validateFieldsToBeNotNull();
+        if (this.getId() == null) throw new IllegalStateException("Can not update a record without ID specified");
+    }
+
+    private void validateToDelete() {
+        if (this.getId() == null) throw new IllegalStateException("Can not delete a record without ID specified");
+    }
+
+    private void validateFieldsToBeNotNull() {
+        if (this.getLengthInches() == null)
```
npathai: I prefer not allowing the invariant to be created in the first place. This validation of `non-nullability` should be in the respective setter methods. In update methods we should validate two or more fields that are dependent on each other.

active-record/src/main/java/com/iluwatar/active/record/SpellCastException.java
```diff
@@ -0,0 +1,17 @@
+package com.iluwatar.active.record;
+
+/**
+ * Describes the exception that is raised when a spell can not be case for some reason.
```
npathai: Typo: `case`

active-record/src/main/java/com/iluwatar/active/record/MagicWand.java
```diff
@@ -0,0 +1,264 @@
...
+ * from which it derives its materials. Moreover, each wand, from the moment it finds its ideal owner, will begin
+ * to learn from and teach its human partner." @ Garrick Ollivander
+ * <p/>
+ * Created by Stephen Lazarionok.
+ */
+public class MagicWand {
+
+    private Long id;
+
+    private Double lengthInches;
```
npathai: Why not `double`? Any reason for choosing boxed type?

active-record/src/main/java/com/iluwatar/active/record/MagicWand.java
```diff
@@ -0,0 +1,264 @@
...
+     * ****** Data access methods
+     * ***********************************
+     */
+
+    private static final String SELECT_SQL = "select * from wand where id = ?";
+    private static final String DELETE_SQL = "delete from wand where id = ?";
+    private static final String UPDATE_SQL = "update wand set length_inches = ?, wood = ?, core = ? where id = ?";
+    private static final String CREATE_SQL = "insert into wand values(?, ?, ?, ?)";
+
+    static {
```
npathai: Doing this makes this tightly coupled with H2, which is fine for this demo example. But we should document it somewhere that in real applications it is not suitable and the class should be multi-tenant, which can be achieved by passing `Connection` object in `save()`, `update()` and other methods that touch database.

active-record/src/main/java/com/iluwatar/active/record/MagicWand.java
```diff
@@ -0,0 +1,264 @@
...
+        catch (final SQLException | ClassNotFoundException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    private static Connection getConnection() throws SQLException {
+        return DriverManager.getConnection("jdbc:h2:~/test");
+    }
+
+    public long save() {
```
npathai: It will be good if we document all public methods.

active-record/src/test/java/MagicWandSpecification.java
```diff
@@ -0,0 +1,60 @@
...
+import org.junit.Assert;
+import org.junit.Test;
+
+/**
+ * Created by Stephen Lazarionok.
+ */
+public class MagicWandSpecification {
+
+    @Test
+    public void shouldCastFireball_IfMaigPower10AndMore() throws SpellCastException {
```
npathai: typo: `Maig` in test method name; should be `Magic`

active-record/src/test/java/MagicWandSpecification.java
```diff
@@ -0,0 +1,60 @@
...
+        final MagicWand wand = new MagicWand();
+        wand.setWood(WandWoodType.WINE);
+        wand.setCore(WandCoreType.DRAGON_HEARTSTRING);
+        wand.setLengthInches(20.5d);
+
+        Assert.assertTrue(wand.calculateMagicPower() == 20.5d);
+        wand.castLighting();
+    }
+
+    @Test(expected = SpellCastException.class)
```
npathai: A better way for testing exceptions is to use `ExpectedException` rule.

A mockup
```java
class Test {
  @Rule public ExpectedException exception = ExpectedException.none();

  @Test
  public void testMethodThatThrowsException() throws SpellCastException {
  
    // Some set up code
    
    exception.expect(SpellCastException.class);
    exception.expectMessage("Can not cast fireball! ...");

    // method that is expected to throw exception
    wand.castFireball();
  }
}

File: active-record/src/test/java/MagicWandSpecification.java [hunk 11]
```diff
@@ -0,0 +1,60 @@
...
+
+    @Test
+    public void shouldCastFireball_IfMaigPower10AndMore() throws SpellCastException {
+
+        final MagicWand wand = new MagicWand();
+        wand.setWood(WandWoodType.WINE);
+        wand.setCore(WandCoreType.DRAGON_HEARTSTRING);
+        wand.setLengthInches(10.0d);
+
+        Assert.assertTrue(wand.calculateMagicPower() == 10.0d);
```
npathai: Use static import

npathai:
We should document disadvantages one of them is violation of SRP, the object knows how it is going to be persisted. How do we manage transaction with Active record? ActiceJDBC does it, but I haven't got time to go in depth. Something about transaction should be said because some real applications rely heavily on transactions.

active-record/src/main/java/com/iluwatar/active/record/MagicWand.java
stephen-lazarionok: Fixed

active-record/src/main/java/com/iluwatar/active/record/MagicWand.java
stephen-lazarionok: Fixed

active-record/src/main/java/com/iluwatar/active/record/SpellCastException.java
stephen-lazarionok: Fixed

active-record/src/main/java/com/iluwatar/active/record/MagicWand.java
stephen-lazarionok: It's necessary to perform checks even though none of setters were invoked. This is why I have this validation.

active-record/src/main/java/com/iluwatar/active/record/MagicWand.java
stephen-lazarionok: I extracted the DB logic into a separate class to achieve low coupling. Passing Connection to save and update methods is a bad approach as it breaks encapsulation.

active-record/src/main/java/com/iluwatar/active/record/MagicWand.java
stephen-lazarionok: Documented however I think methods like save(), update(), delete() are self-documented already and having javadocs are verbose.

active-record/src/test/java/MagicWandSpecification.java
stephen-lazarionok: Fixed.

active-record/src/test/java/MagicWandSpecification.java
stephen-lazarionok: Changed to ExpectedException. But why do you think this is the better way?

active-record/src/main/java/com/iluwatar/active/record/MagicWand.java
stephen-lazarionok: Fixed

active-record/src/test/java/MagicWandSpecification.java
stephen-lazarionok: Changed to static import. However why do you recommend using static methods instead of normal ones?

stephen-lazarionok:
It's disputable point about SRP violation as we just put all the logic including domain and persistence related to some record into one class.
Transactions management is orthogonal topic that is not really connected to patterns. I think it's not worth mentioning it.

active-record/src/main/java/com/iluwatar/active/record/DB.java
```diff
@@ -0,0 +1,38 @@
...
+        DeleteDbFiles.execute("~", "test", true);
+        try {
+            Class.forName("org.h2.Driver");
+            final Connection connection = getConnection();
+            final Statement statement = connection.createStatement();
+            statement.execute("create table wand(id BIGINT primary key, length_inches REAL, wood varchar(100), core varchar(100))");
+            statement.close();
+            connection.close();
+        }
+        catch (final SQLException | ClassNotFoundException e) {
```
npathai: Style nit: `catch` should be in the line of closing `try` curly brace.

```java
try {
  ...
} catch (Exception ex) {
  ...
}

I personally prefer Google style guide. I will also raise this point in chat room. What do you prefer here?

File: active-record/src/main/java/com/iluwatar/active/record/DB.java [hunk 13]
```diff
@@ -0,0 +1,38 @@
+package com.iluwatar.active.record;
+
+import org.h2.tools.DeleteDbFiles;
+
+import java.sql.Connection;
+import java.sql.DriverManager;
+import java.sql.SQLException;
+import java.sql.Statement;
+
+/**
```
npathai: Add javadoc explaining why this class is important. Mainly that it allows making the solution Multi-tenant. Switching DB will not require code change in `MagicWand`.

active-record/src/main/java/com/iluwatar/active/record/MagicWand.java
```diff
@@ -0,0 +1,261 @@
...
+
+            final long id = System.currentTimeMillis();
+            ps.setLong(1, id);
+            ps.setDouble(2, getLengthInches());
+            ps.setString(3, getWood().name());
+            ps.setString(4, getCore().name());
+
+            ps.execute();
+
+            ps.close();
```
npathai: I am nitpicking here but it is not a good idea to close connections in `try` block, they must be in `finally` block. This does not matter in our small example. But I would prefer we don't violate best practices.

active-record/src/main/java/com/iluwatar/active/record/MagicWand.java
npathai: Is there a specific reason you have not asked for mandatory arguments such as length, core and wood in constructor? Creating a wand without mandatory arguments is what allows creation of those invariants.
Another point that I was thinking about was can the wood and type of wand be altered post creation? Perhaps we can have a color property which can be altered post creation of wand.

active-record/src/main/java/com/iluwatar/active/record/MagicWand.java
npathai: The DB approach is ok, but in any way what we have now is tight coupling with DB lurking around with Magic wand class and the dependency isn't visible in any way as you have static method and you reach for connection in global space. What will happen if you created a `MagicWand` and called `save` on it but DB class needed some initialization? Presently it seems fine in `static` block but as the initialization logic of DB gets complicated things get out of hand. In any case it is a leaky abstraction. Present implementation is also fine with me though.

active-record/src/test/java/MagicWandSpecification.java
npathai: I will point you to a link: http://blog.codeleak.pl/2014/03/junit-expectedexception-rule-beyond.html

active-record/src/test/java/MagicWandSpecification.java
npathai: There aren't but using static import here reduces verbosity and looks readable. Again regarding code guideline I will raise this point in chat.

npathai:
Regarding Transaction management, though it is an orthogonal topic but it will affect the implementation. Imagine what would we do if the example was about `Account` and we had a transaction of money transfer. Then the `withdraw()` method of `Account` would not have been able to commit and close the connection. It was because of this that I was wondering if we should mention about transactions.
We have several patterns related to persistence: #55 , #34 #45 , so it is important to compare them and thinking and documenting where each of these patterns shine and have limitations will make things clearer in readers mind. Let me know your comments on this.
@iluwatar Maybe you can pitch in here.

active-record/src/main/java/com/iluwatar/active/record/DB.java
stephen-lazarionok: Fixed. I do not have any particular preferences :)

active-record/src/main/java/com/iluwatar/active/record/DB.java
stephen-lazarionok: Add the documentation to the DB class

active-record/src/main/java/com/iluwatar/active/record/MagicWand.java
stephen-lazarionok: Nice catch! Fixed.

npathai:
@stephen-lazarionok I have a few doubts and need your clarification on them. Please go through my comments and respond. After that I think we are good to go.

active-record/src/main/java/com/iluwatar/active/record/MagicWand.java
stephen-lazarionok: The alternative here is to use something like Spring IoC, make DB as a single bean and wrap it with the interface. However it will make an example more complicated. I want to keep it is very simple. The same approach was used by Martin Fowler to show Active Record pattern.

active-record/src/test/java/MagicWandSpecification.java
stephen-lazarionok: Okay, got it. Still think that the annotation based approach is more readable and simpler even though will try this approach with rules :)

active-record/src/main/java/com/iluwatar/active/record/MagicWand.java
stephen-lazarionok: I am not a big fan of moving all the mandatory fields to constructor as then we end up supporting something like SomeObj(String, String, Int, String, Double, Enum). In addition to this we will mix up the different logic. Probably it's okay to create a wand without a core but it's not okay to save it to some DB due to some business rules. Because of that I used contextual validation - http://martinfowler.com/bliki/ContextualValidation.html.

stephen-lazarionok:
@npathai I commented on your points. I guess it's worth discussing the point about transactions and patterns related to DBs. I checked out other patterns implemented earlier and did not find any references to transactions. My opinion is still that transactions is a separate topic and we should not mix it up with patterns. We could add this information later if needed.

active-record/src/main/java/com/iluwatar/active/record/MagicWand.java
npathai: @stephen-lazarionok A `MagicWand` will not be created only to be saved in database. What if a wand is created to cast a fireball? In that scenario we allowed an invariant that is core should not be `null ` and when `castFireball()` will be called it will say "Not enough magic power" but actually the wand is not yet built. I hope you understand my concern. But again these are all nitpicks. We can even utilize `Builder` based #184 for creating a wand if constructor is getting too long. But let's not over-engineer :smiley:

active-record/src/main/java/com/iluwatar/active/record/MagicWand.java
npathai: @stephen-lazarionok Yes I hear you. I am also willing to keep the example simple. We have introduced consequences section in our pattern description with #205 . And I feel discussing these points will help us in the learning process and we can get something out of it for consequences section as well. Discussions on design patterns tend to get opinionated but hey that's the fun part. Ain't it? As far as the implementation is concerned I am happy with it.

npathai:
@stephen-lazarionok Yes as far as the previous patterns are concerned, they were created by very few people maybe by @iluwatar and now we have a team. So we can improve upon many fronts. We will be happy to hear your thoughts on transactions on previous patterns as well it's something worth mentioning. Also in future we will be visiting previous patterns and improve upon them.

npathai:
@iluwatar LGTM. If you are satisfied with implementation then we can proceed with README and web site changes.

npathai:
Any updates @stephen-lazarionok ?

stephen-lazarionok:
Sorry, was busy other stuff on my projects. Will take care of that this weekend.

stephen-lazarionok:
Had to take care of some urgent stuff this weekend, will do tomorrow.

npathai:
@stephen-lazarionok Not a problem :+1:

stephen-lazarionok:
Format code according to Google Style Guide     ---> **done**
 Prepare diagram using Eclipse ObjectAid plugin. ---> Need help to generate the diagram
 Implement README and website changes          ---> **done** (todo: just need to add the link to the diagram)

npathai:
@stephen-lazarionok How can I help? Did you face any issue while generating the diagram?

stephen-lazarionok:
Will install Eclipse and fix soon. Sorry, do not have enough time as I am in the middle of my trip and also have stuff to do on my projects.

npathai:
Not an issue :) take ur time..

iluwatar:
This pull request made no progress during the last 30 days, so I will close it.

PR #220, opened 2015-08-18
Working back to top button.

mafagafogigante:
Needs some peer review in the JS part. I used a cleaner solution and removed useless (chinese?) comments.

It wasn't working in Firefox 40 for me, so I guess it is a pretty big deal.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,12 +1,14 @@
 # Design pattern samples in Java.
 
-This is the branch for the pages
+This is the branch for the pages.
 
 Link for Website: http://iluwatar.github.io/java-design-patterns/
 
 ## How to update the website reference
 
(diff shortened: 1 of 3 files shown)


noobxgockel:
Well, this still seems like a better solution until someone finds the time to look harder for a "smooth" scroll...

mafagafogigante:
I would like to make it clear if I failed to do so with the commit message. Smooth scrolling was also **fixed** (I considered it an issue, even if it worked perfectly). This is reference material, it can be pretty without being bloated and animations are bloat.

mafagafogigante:
Anyway, the important part is that this is a fix and should be merged. The animations (or their absence) can be discussed in the future.

Thanks for merging.

PR #223, opened 2015-08-20
Junit Test case added for Template method pattern

sklrsn:
Junit Test case added for Template method pattern


iluwatar:
I find this test no so useful, since it does exactly the same thing as https://github.com/iluwatar/java-design-patterns/blob/master/template-method/src/test/java/com/iluwatar/templatemethod/AppTest.java

PR #225, opened 2015-08-21
#184 Fluent Interface pattern

hannespernpeintner:


npathai:
@hannespernpeintner Did you get the lazy meaning? You can try implementing if you want. You know it never hurts to have something extra.

hannespernpeintner:
Yes i got it. I'll implement a lazy version as well :)

hannespernpeintner:
Okay, took me more time than expected because the lazy implementation is more complex than the eager one - but I added a lazy implementation and extracted an interface.

Would be nice to hear your thoughts :)

npathai:
@hannespernpeintner I have been a bit busy lately. Will be sending initial comments soon.

hannespernpeintner:
Hey, what about your feedback? :)

npathai:
@hannespernpeintner Give me a day or two :)

fluentinterface/src/main/java/com/iluwatar/fluentinterface/fluentiterable/FluentIterable.java
```diff
@@ -0,0 +1,81 @@
...
+/**
+ * The FluentIterable is a more convenient implementation of the common iterable interface based
+ * on the fluent interface design pattern.
+ * This interface defines common operations, but
+ * doesn't aim to be complete. It was inspired by Guava's com.google.common.collect.FluentIterable.
+ * @param <TYPE> is the class of objects the iterable contains
+ */
+public interface FluentIterable<TYPE> extends Iterable<TYPE> {
+
+    /**
```
npathai: Other way: "Filters the contents of `Iterable` using the given predicate, leaving only the ones which _satisfy_ the predicate."

fluentinterface/src/main/java/com/iluwatar/fluentinterface/fluentiterable/FluentIterable.java
```diff
@@ -0,0 +1,81 @@
...
+     *                  is negative, the tested object is removed by the iterator.
+     * @return a filtered FluentIterable
+     */
+    FluentIterable<TYPE> filter(Predicate<? super TYPE> predicate);
+
+    /**
+     * Uses the Iterable interface's forEach method to apply a given function
+     * for each object of the iterator. This is a terminating operation.
+     */
+    void forEachDo(Consumer<? super TYPE> action);
```
npathai: Java 8 `forEach` method would suffice. We dont need to provide other method.

fluentinterface/src/main/java/com/iluwatar/fluentinterface/fluentiterable/FluentIterable.java
```diff
@@ -0,0 +1,81 @@
...
+
+    /**
```
npathai: Javadoc: Returns an `Optional` containing the first element of this iterable if present, else returns `Optional.empty()`.

fluentinterface/src/main/java/com/iluwatar/fluentinterface/fluentiterable/FluentIterable.java
```diff
@@ -0,0 +1,81 @@
...
+     * Evaluates the iteration and returns the first element. This is a terminating operation.
+     * @return the first element after the iteration is evaluated
+     */
+    Optional<TYPE> first();
+
+    /**
```
npathai: Javadoc: `Evaluates` is not a correct representation in interface because in case of Lazy iterable the evaluation will only occur on termination operation.

Javadoc: `iteration` in javadoc should be changed to `Iterable`

fluentinterface/src/main/java/com/iluwatar/fluentinterface/fluentiterable/FluentIterable.java
```diff
@@ -0,0 +1,81 @@
...
+    FluentIterable<TYPE> last(int count);
+
+    /**
+     * Transforms this FluentIterable into a new one containing objects of the type NEW_TYPE.
+     * @param function a function that transforms an instance of TYPE into an instance of NEW_TYPE
+     * @param <NEW_TYPE> the target type of the transformation
+     * @return a new FluentIterable of the new type
+     */
+    <NEW_TYPE> FluentIterable<NEW_TYPE> map(Function<? super TYPE, NEW_TYPE> function);
+    List<TYPE> asList();
```
npathai: Missed javadoc for `asList()`

fluentinterface/src/main/java/com/iluwatar/fluentinterface/fluentiterable/lazy/DecoratingIterator.java
```diff
@@ -0,0 +1,53 @@
...
+        this.fromIterator = fromIterator;
+    }
+
+    /**
+     * Precomputes and caches the next element of the iteration.
+     * @return true if a next element is available
+     */
+    @Override
+    public final boolean hasNext() {
+        next = computeNext();
```
npathai: The `hasNext()` should not compute next element if the element is not consumed using `next()`.

fluentinterface/src/main/java/com/iluwatar/fluentinterface/fluentiterable/lazy/DecoratingIterator.java
```diff
@@ -0,0 +1,53 @@
...
+        return next != null;
+    }
+
+    /**
+     * Returns the next element of the iteration. This implementation caches it.
+     * If no next element is cached, it is computed.
+     * @return the next element obf the iteration
+     */
+    @Override
+    public final TYPE next() {
```
npathai:
```java
if (next == null) {
  return fromIterator.next();
} else {
  final TYPE result = next;
  next = null;
  return result;
}
```
I find this better to read.

fluentinterface/src/main/java/com/iluwatar/fluentinterface/fluentiterable/lazy/DecoratingIterator.java
```diff
@@ -0,0 +1,53 @@
...
+     * @return true if a next element is available
+     */
+    @Override
+    public final boolean hasNext() {
+        next = computeNext();
+        return next != null;
+    }
+
+    /**
+     * Returns the next element of the iteration. This implementation caches it.
```
npathai: This implementation _does not_ cache. It _returns_ cached element if present, else computes it (does not cache).

fluentinterface/src/main/java/com/iluwatar/fluentinterface/fluentiterable/lazy/DecoratingIterator.java
```diff
@@ -0,0 +1,53 @@
...
+        result = (result == null ? fromIterator.next() : result);
+        return result;
+    }
+
+    /**
+     * Computes the next object of the iteration. Can be implemented to
+     * realize custom behaviour for an iteration process.
+     * @return
+     */
+    public abstract TYPE computeNext();
```
npathai: Nitpick: this implementation will not work if `null` can represent a valid computation value. If the underlying iterator returns `null`, it can mean two things end of iteration or actual `null`. Maybe we should not change the present behavior, just document it.
Javadoc: `null` is considered as end of data should be documented.

fluentinterface/src/main/java/com/iluwatar/fluentinterface/fluentiterable/lazy/LazyFluentIterable.java
```diff
@@ -0,0 +1,236 @@
...
+     */
+    @Override
+    public FluentIterable<TYPE> filter(Predicate<? super TYPE> predicate) {
+        return new LazyFluentIterable<TYPE>() {
+            @Override
+            public Iterator<TYPE> iterator() {
+                return new DecoratingIterator<TYPE>(iterable.iterator()) {
+                    @Override
+                    public TYPE computeNext() {
+                        while(true) {
```
npathai: why not `while(fromIterator.hasNext())`?

fluentinterface/src/main/java/com/iluwatar/fluentinterface/fluentiterable/lazy/LazyFluentIterable.java
```diff
@@ -0,0 +1,236 @@
...
+        };
+    }
+
+    /**
+     * Uses the Iterable interface's forEach method to apply a given function
+     * for each object of the iterator. Is a terminating operation.
+     * @param action the action for each object
+     */
+    @Override
+    public void forEachDo(Consumer<? super TYPE> action) {
```
npathai: I think we don't need this.

fluentinterface/src/main/java/com/iluwatar/fluentinterface/fluentiterable/lazy/LazyFluentIterable.java
```diff
@@ -0,0 +1,236 @@
...
+    }
+
+    /**
+     * Can be used to collect objects from the iteration. Is a terminating operation.
+     * @return an option of the first object of the iteration
+     */
+    @Override
+    public Optional<TYPE> first() {
+        Optional result = Optional.empty();
+        List<TYPE> list = first(1).asList();
```
npathai: How about
```java
Iterable<TYPE> resultIterable = first(1);
return resultIterable.hasNext() ? Optional.of(resultIterable.next()) : Optional.empty();
``` 
We don't need to copy it into list.

fluentinterface/src/main/java/com/iluwatar/fluentinterface/fluentiterable/lazy/LazyFluentIterable.java
```diff
@@ -0,0 +1,236 @@
...
+    public void forEachDo(Consumer<? super TYPE> action) {
+        Iterator<TYPE> newIterator = iterable.iterator();
+        while(newIterator.hasNext()) {
+            action.accept(newIterator.next());
+        }
+    }
+
+    /**
+     * Can be used to collect objects from the iteration. Is a terminating operation.
+     * @return an option of the first object of the iteration
```
npathai: Javadoc: an `Optional` containing first object of this `Iterable`.

fluentinterface/src/main/java/com/iluwatar/fluentinterface/fluentiterable/lazy/LazyFluentIterable.java
```diff
@@ -0,0 +1,236 @@
...
+        List<TYPE> list = first(1).asList();
+        if(!list.isEmpty()) {
+            result = Optional.of(list.get(0));
+        }
+
+        return result;
+    }
+
+    /**
+     * Can be used to collect objects from the iteration. Is a terminating operation.
```
npathai: Javadoc: This is not a terminating operation.

fluentinterface/src/main/java/com/iluwatar/fluentinterface/fluentiterable/lazy/LazyFluentIterable.java
```diff
@@ -0,0 +1,236 @@
...
+                        return null;
+                    }
+                };
+            }
+        };
+    }
+
+    /**
+     * Can be used to collect objects from the iteration. Is a terminating operation.
+     * @return an option of the last object of the iteration
```
npathai: Javadoc: an `Optional` containing last object of this Iterable.

fluentinterface/src/main/java/com/iluwatar/fluentinterface/fluentiterable/lazy/LazyFluentIterable.java
```diff
@@ -0,0 +1,236 @@
...
+     */
+    @Override
+    public Optional<TYPE> last() {
+        Optional result = Optional.empty();
+        List<TYPE> list = last(1).asList();
```
npathai: Similar change as `first(count)`.

fluentinterface/src/main/java/com/iluwatar/fluentinterface/fluentiterable/lazy/LazyFluentIterable.java
```diff
@@ -0,0 +1,236 @@
...
+        return result;
+    }
+
+    /**
+     * Can be used to collect objects from the iteration. Is a terminating operation.
+     * @param count defines the number of objects to return
+     * @return the same FluentIterable with a collection decimated to a maximum of 'count' last objects
+     */
+    @Override
+    public FluentIterable<TYPE> last(int count) {return new LazyFluentIterable<TYPE>() {
```
npathai: Do you think we should support this operation? This is too _memory intensive_. An iterable can hypothetically represent huge stream of data and to store that in memory would not be advisable. If we decide to keep this then we need to document its memory footprint.

fluentinterface/src/main/java/com/iluwatar/fluentinterface/fluentiterable/lazy/LazyFluentIterable.java
```diff
@@ -0,0 +1,236 @@
...
+    @Override
+    public <NEW_TYPE> FluentIterable<NEW_TYPE> map(Function<? super TYPE, NEW_TYPE> function) {
+        return new LazyFluentIterable<NEW_TYPE>() {
+            @Override
+            public Iterator<NEW_TYPE> iterator() {
+                return new DecoratingIterator<NEW_TYPE>(null) {
+                    Iterator<TYPE> oldTypeIterator = iterable.iterator();
+                    @Override
+                    public NEW_TYPE computeNext() {
+                        while(true) {
```
npathai: Logic: 
`while(oldTypeIterator.hasNext())`

fluentinterface/src/main/java/com/iluwatar/fluentinterface/fluentiterable/simple/SimpleFluentIterable.java
```diff
@@ -0,0 +1,194 @@
...
+ */
+public class SimpleFluentIterable<TYPE> implements FluentIterable<TYPE> {
+
+    private final Iterable<TYPE> iterable;
+
+    /**
+     * This constructor creates a copy of a given iterable's contents.
+     * @param iterable the iterable this interface copies to work on.
+     */
+    protected SimpleFluentIterable(Iterable<TYPE> iterable) {
```
npathai: Doubt: I couldn't figure out the reason for constructor being `protected`.

fluentinterface/src/main/java/com/iluwatar/fluentinterface/fluentiterable/simple/SimpleFluentIterable.java
```diff
@@ -0,0 +1,194 @@
...
+        iterable.forEach(action);
+    }
+
+    /**
+     * Can be used to collect objects from the iteration. Is a terminating operation.
+     * @return an option of the first object of the iteration
+     */
+    @Override
+    public final Optional<TYPE> first() {
+        List<TYPE> list = first(1).asList();
```
npathai: Logic: Similar change to `LazyFluentIterable#first()`

npathai:
@hannespernpeintner I am done with first review. Awaiting your response.

fluentinterface/src/main/java/com/iluwatar/fluentinterface/fluentiterable/FluentIterable.java
hannespernpeintner: Why is this not a terminating operation?
(I changed iteration to Iterable everywhere.)

fluentinterface/src/main/java/com/iluwatar/fluentinterface/fluentiterable/lazy/DecoratingIterator.java
hannespernpeintner: I can't see a way how to not request the next element, in order to figure out if there's a next one or not with stacked iterators. A fluent iterator with an additional action has to apply this action onto the next element, so it has to be computed. I think that's also the way guava works internally, or did I miss the point?

fluentinterface/src/main/java/com/iluwatar/fluentinterface/fluentiterable/lazy/LazyFluentIterable.java
hannespernpeintner: I think it would be nice to have it, because that's probably one thing people have problems to implement by themselves - regardless if the operation would be cheap or expensive. I added additional documentation instead.

fluentinterface/src/main/java/com/iluwatar/fluentinterface/fluentiterable/simple/SimpleFluentIterable.java
hannespernpeintner: * The actual reason is homogenity: For some implementations of the FluentIterable, more than one constructor is required (for example LazyFluentIterable). In order to prevent calling constructors of FluentIterable implementations directly, they are provided via static factory methods on the implementing class.
In this case, a public constructor could be provided, but then again, it should be provided for the lazy implementation as well - and for the lazy class, there are currently two constructors of which at least one has to be protected (the parameterless one). Would it be better to switch to public constructors?
* It would be possible to use an incoming _iterator_ - but it would be problematic if an _iterable_ is provided, because this would be misleading for the user of our implementation. If an iterator was provided, the user could expect that no defensive copy was made, but it would have made some of the example methods more complicated, I think. Do you agree?

hannespernpeintner:
Thank you very much for your excellent feedback. I corrected everything you stated (I think) - that's the reason I passed on replying to some of your comments. Some points probably need further discussion,  so I replied to them directly.

fluentinterface/src/main/java/com/iluwatar/fluentinterface/fluentiterable/FluentIterable.java
npathai: Because `first(count)` is not evaluated until a terminating function like `forEach()`, `max(), `min()` is called (For Lazy implementation).
Example:
```java
Optional<Integer> max = FluentIterable.from(1, 2, 3, -1, -2, -3)
                                      .first(4) // we don't need to evaluate first 4 until a terminate call is made
                                      .max();
```
I hope you got the point. `first()` is a terminating call and `first(count)` is not.

fluentinterface/src/main/java/com/iluwatar/fluentinterface/fluentiterable/lazy/DecoratingIterator.java
npathai: Guava `AbstractIterable` has a `state` property which changes to `READY` when `computeNext()` returns a element. And then on consecutive calls to `hasNext()`, if `state` is `READY` then it returns `true` and does not compute next element.
Like this:
```java
public boolean hasNext() {
		switch (state) {
		case READY:
			return true;
		case DONE:
			return false;
		default:
		}
		return tryComputeNext();
	}
```

fluentinterface/src/main/java/com/iluwatar/fluentinterface/fluentiterable/lazy/LazyFluentIterable.java
```diff
@@ -0,0 +1,211 @@
...
+        @Override
+        public Iterator<TYPE> iterator() {
+            return new DecoratingIterator<TYPE>(iterable.iterator()) {
+                int currentIndex = 0;
+
+                @Override
+                public TYPE computeNext() {
+                    List<TYPE> list = new ArrayList<>();
+
+                    Iterator<TYPE> newIterator = iterable.iterator();
```
npathai: We don't need to populate the list on each `computeNext()`. You can cache it on first call and then utilize it. If you have doubts then I am implementing it, will give you sample.

fluentinterface/src/main/java/com/iluwatar/fluentinterface/fluentiterable/simple/SimpleFluentIterable.java
npathai: Feedback
 - I got it. We can have `protected` constructors.
 - I think we should use incoming `Iterable` itself because I would like if the original collection is altered so that I don't have to pay extra cost. In case user does not want to mutate original collection then we can provide other static factory method `fromCopyOf(Iterable)`. Then it is upto the user. I am leaning towards mutating original because that's what `LazyFluentIterable` does, so it becomes homogeneous :smiley:

npathai:
Summary of review:
 - Logic: I have some doubts about implementation of `last(count)` in both versions. Will be giving feedback on that after trying something myself.
 - Suggestion: It is difficult to review the changes undertaken so I suggest you to leave comments on the points incorporated as well. Something like `Done`! will be quite helpful.
 - Suggestion: We can document that `SimpleFluentIterable` is too costly to be utilized in real applications.
 - Doubt: Do we need `forEach()` method in our `FluentIterable`s as they already extend `Iterable` and there is default implementation in `interface`.

fluentinterface/src/main/java/com/iluwatar/fluentinterface/fluentiterable/FluentIterable.java
hannespernpeintner: Done.

fluentinterface/src/main/java/com/iluwatar/fluentinterface/fluentiterable/FluentIterable.java
hannespernpeintner: Removed.

fluentinterface/src/main/java/com/iluwatar/fluentinterface/fluentiterable/FluentIterable.java
hannespernpeintner: Done.

fluentinterface/src/main/java/com/iluwatar/fluentinterface/fluentiterable/FluentIterable.java
hannespernpeintner: Ah okay I got you. Your comment was added to "first()" instead of "first(count)" - that's what confused me.

Done.

fluentinterface/src/main/java/com/iluwatar/fluentinterface/fluentiterable/FluentIterable.java
hannespernpeintner: Added.

fluentinterface/src/main/java/com/iluwatar/fluentinterface/fluentiterable/lazy/DecoratingIterator.java
hannespernpeintner: Do you think I should implement this as well with an internal state?

fluentinterface/src/main/java/com/iluwatar/fluentinterface/fluentiterable/lazy/DecoratingIterator.java
hannespernpeintner: Done.

fluentinterface/src/main/java/com/iluwatar/fluentinterface/fluentiterable/lazy/DecoratingIterator.java
hannespernpeintner: Done.

fluentinterface/src/main/java/com/iluwatar/fluentinterface/fluentiterable/lazy/DecoratingIterator.java
hannespernpeintner: Added documentation.

fluentinterface/src/main/java/com/iluwatar/fluentinterface/fluentiterable/lazy/LazyFluentIterable.java
hannespernpeintner: Good question. Changed.

fluentinterface/src/main/java/com/iluwatar/fluentinterface/fluentiterable/lazy/LazyFluentIterable.java
hannespernpeintner: Removed.

fluentinterface/src/main/java/com/iluwatar/fluentinterface/fluentiterable/lazy/LazyFluentIterable.java
hannespernpeintner: hasNext has to be called on the Iterator, not the Iterable, but I changed it similar to your example. Thanks,

fluentinterface/src/main/java/com/iluwatar/fluentinterface/fluentiterable/lazy/LazyFluentIterable.java
hannespernpeintner: Done.

fluentinterface/src/main/java/com/iluwatar/fluentinterface/fluentiterable/lazy/LazyFluentIterable.java
hannespernpeintner: Done.

fluentinterface/src/main/java/com/iluwatar/fluentinterface/fluentiterable/lazy/LazyFluentIterable.java
hannespernpeintner: Done.

fluentinterface/src/main/java/com/iluwatar/fluentinterface/fluentiterable/lazy/LazyFluentIterable.java
hannespernpeintner: Done.

fluentinterface/src/main/java/com/iluwatar/fluentinterface/fluentiterable/lazy/LazyFluentIterable.java
hannespernpeintner: Done.

fluentinterface/src/main/java/com/iluwatar/fluentinterface/fluentiterable/simple/SimpleFluentIterable.java
hannespernpeintner: I added a copyFactoryMethod for the simple implementation. I wait before I push the commit because one thing has to be discussed first: I'm not sure if I implemented the lazy version correct/as you expected: It doesn't touch the incoming collection, because on termination a new iterator of the incoming iteratable is requested, used for iteration but no element is ever removed. Otherwise, one would have to handle ConcurrentModificationException or OperationNotSupportedException and I'm currently not quite sure if this is easy doable.

fluentinterface/src/main/java/com/iluwatar/fluentinterface/fluentiterable/lazy/LazyFluentIterable.java
hannespernpeintner: Yes you're right. I changed it.

hannespernpeintner:
* Thank you for your effort. What are your doubts about: performance, if they're necessary, if they are wrong?
* I added a comment for all comments of yours.
* Should this be added to the class documentation, or what's the best place to add it?
* My first implementation allowed to embed it as a non-terminating operation in a fluent usage, but later, I changed it. Now I removed it, we don't need it as it is now.

Thank you so far, I'm looking forward to your next comments :)

fluentinterface/src/main/java/com/iluwatar/fluentinterface/fluentiterable/lazy/DecoratingIterator.java
npathai: Not necessarily. We can document that it is a toy implementation and does not take care of all scenarios such as consecutive `hasNext()` calls. I raised the point for your understanding :smiley:

fluentinterface/src/main/java/com/iluwatar/fluentinterface/fluentiterable/lazy/LazyFluentIterable.java
npathai: Nice catch. Typo.

fluentinterface/src/main/java/com/iluwatar/fluentinterface/fluentiterable/simple/SimpleFluentIterable.java
npathai: I totally messed it up in my mind. You are correct that lazy does not mutate original collection. We can have your implementation of Simple. Sorry for that.

npathai:
- Yes about performance.
 - Great.
 - Yes class documentation would be perfect!
 - Great.

I am moving towards final review. Will provide comments soon.

fluentinterface/src/main/java/com/iluwatar/fluentinterface/fluentiterable/lazy/DecoratingIterator.java
hannespernpeintner: I added a line to the class documentation.

fluentinterface/src/main/java/com/iluwatar/fluentinterface/fluentiterable/simple/SimpleFluentIterable.java
hannespernpeintner: Absolutely no problem :). So this is done.

hannespernpeintner:
I ammended documentation changes. Nothing left to do currently, thank you for your help, I'm waiting for further responses :)

npathai:
LGTM you can move towards other tasks to complete this PR.

Tasks:
 - [X] Implementation
 - [X] Documentation
 - [ ] Code formatting according to Google Java guide
 - [ ] Verify coverage
 - [ ] Create class diagram
 - [ ] Provide website content - visit wiki for steps

hannespernpeintner:
The latest commit should complete the PR. However, I can't find a guideline to your code coverage rules. Currently, the coverage is 91.5%.

npathai:
@hannespernpeintner Its good enough :+1: I will bring this point up on chat room and soon we will add the guideline on coverage ratio.

fluentinterface/src/main/java/com/iluwatar/fluentinterface/fluentiterable/lazy/LazyFluentIterable.java
```diff
@@ -0,0 +1,231 @@
...
+   * @return the same FluentIterable with a collection decimated to a maximum of 'count' last
+   *         objects
+   */
+  @Override
+  public FluentIterable<TYPE> last(int count) {
+    return new LazyFluentIterable<TYPE>() {
+      @Override
+      public Iterator<TYPE> iterator() {
+        return new DecoratingIterator<TYPE>(iterable.iterator()) {
+          public int stopIndex;
```
npathai: Why `public`?

npathai:
@hannespernpeintner You have not created an `index.md` file as suggested in [How to contribute](https://github.com/iluwatar/java-design-patterns/wiki/1.-How-to-contribute) section. Your branch is far too outdated, you need to merge the changes from `iluwatar/master` into your fork/master branch. We have removed the patterns from README and created website pages. So you need to incorporate those changes.

hannespernpeintner:
Ah, this totally bypassed me. I rebased and added the description. Anything else (the ci seems to fail)? :)

fluentinterface/src/main/java/com/iluwatar/fluentinterface/fluentiterable/lazy/LazyFluentIterable.java
hannespernpeintner: Was a mistake, made it private.

fluentinterface/index.md
```diff
@@ -0,0 +1,24 @@
...
+Fluent interfaces are sometimes confused with the Builder pattern, because they share method chaining and a fluent usage. However, fluent interfaces are not primarily used to create objects.
+
+![Fluent Interface](./etc/fluentinterface.png "Fluent Interface")
+
+**Applicability:** Use the Fluent Interface pattern when
+
+* you provide an API that would benefit from a DSL-like usage
+* you have objects that are difficult to configure or use
+
+**Real world examples:**
```
npathai: Add `Credits` section.

fluentinterface/index.md
```diff
@@ -0,0 +1,24 @@
+---
+layout: pattern
+title: Fluent Interface
+folder: fluentinterface
+permalink: /patterns/fluentinterface/
+categories: Architectural
+tags: Java
+---
+
+**Intent:** A fluent interface provides an easy-readable, flowing interface, that often mimics a domain specific language. This is especially useful when the users of your API benefit from code, that can nearly be read as human language.
```
npathai: > This is especially useful when the users of your API benefit from code, that can nearly be read as human language.

Can we rephrase this?

fluentinterface/index.md
```diff
@@ -0,0 +1,24 @@
...
+![Fluent Interface](./etc/fluentinterface.png "Fluent Interface")
+
+**Applicability:** Use the Fluent Interface pattern when
+
+* you provide an API that would benefit from a DSL-like usage
+* you have objects that are difficult to configure or use
+
+**Real world examples:**
+
+* [Java 8 Stream API](http://www.oracle.com/technetwork/articles/java/ma14-java-se-8-streams-2177646.html)
```
npathai: JOOQ is also a good real world example. It represents `Query Builder` and is a well known fluent interface implementation in various languages.

fluentinterface/src/main/java/com/iluwatar/fluentinterface/App.java
```diff
@@ -0,0 +1,110 @@
...
+
+import com.iluwatar.fluentinterface.fluentiterable.lazy.LazyFluentIterable;
+import com.iluwatar.fluentinterface.fluentiterable.simple.SimpleFluentIterable;
+
+import java.util.*;
+import java.util.function.Function;
+import java.util.function.Predicate;
+
+import static java.lang.String.valueOf;
+
```
npathai: Add documentation guiding the reader on how to understand this class and pattern. Refer documentation in `App.java` of other patterns.

npathai:
@hannespernpeintner Include `Iterable` in UML diagram. Also add association links between the classes. `App` uses `FluentIterable` and similar.

fluentinterface/index.md
hannespernpeintner: What do you want me to write under the credits section? FluentInterface from Guava is linked in the real world examples and the interface's class documentation states the inspiration from Guava.

fluentinterface/index.md
hannespernpeintner: Of course, if one has a better idea how to express it - english is not my native language ;)

I wanted to express that this kind of interface is most useful for people that really want to have an easy API, because they're not interested further in the topic of your program, but want to utilize it as quickly as possible. Or when the fluent interface design would really be much better then a traditional approach.

fluentinterface/index.md
hannespernpeintner: Added.

fluentinterface/src/main/java/com/iluwatar/fluentinterface/App.java
hannespernpeintner: Added.

hannespernpeintner:
Included. I considered some relationships as unimportant, so I left them out. Everything pushed, what else? :)

fluentinterface/index.md
npathai: Something like: 
> Using this pattern results in code that can be read nearly as human language.

fluentinterface/index.md
npathai: - [Fluent Interface - Martin Fowler](http://www.martinfowler.com/bliki/FluentInterface.html )
 - [Evolutionary architecture and emergent design: Fluent interfaces](http://www.ibm.com/developerworks/library/j-eaed14/)

fluentinterface/index.md
```diff
@@ -0,0 +1,25 @@
...
+title: Fluent Interface
+folder: fluentinterface
+permalink: /patterns/fluentinterface/
+categories: Architectural
+tags: Java
+---
+
+**Intent:** A fluent interface provides an easy-readable, flowing interface, that often mimics a domain specific language. This is especially useful when the users of your API benefit from code, that can nearly be read as human language.
+
+Fluent interfaces are sometimes confused with the Builder pattern, because they share method chaining and a fluent usage. However, fluent interfaces are not primarily used to create objects.
```
npathai: This can also be a part of FAQ?
"What is the difference between FluentInterface and Builder pattern?"

npathai:
@hannespernpeintner I have commented on a few last changes. Once done we can merge it :tada:

fluentinterface/index.md
hannespernpeintner: Changed.

fluentinterface/index.md
hannespernpeintner: Added.

fluentinterface/index.md
hannespernpeintner: Moved it to the FAQ and extended it a bit.

hannespernpeintner:
I corrected them, kindly ask you to have the last look on it :)

PR #234, opened 2015-09-02
Context Object pattern

ekostadinov:


iluwatar:
This PR seems to be incomplete. There is no implementation.

PR #235, opened 2015-09-03
Jekyll does not like enumerations w/o blank lines

simon04:
See on http://iluwatar.github.io/java-design-patterns/patterns/flyweight/ that `*` are not converted to proper unordered lists.


iluwatar:
Thanks!

PR #240, opened 2015-09-04
Improved unit tests so assertions are used.

themoffster:
The test classes printed to the console and didn't actually do any asserts so you could easily misread and think that it worked. By adding unit tests, it's very easy to see if there are any errors.


iluwatar:
Thanks! Keep them coming :)

npathai:
Thanks.

Issue #224, opened 2015-08-21
Reformat existing code using Google Style Guide


npathai:
This issue is related to #221 . With this issue we will re-format existing implementation to confirm to Google Style guide.

iluwatar:
https://github.com/google/styleguide/blob/gh-pages/eclipse-java-google-style.xml does not seem to work in Eclipse Mars.

npathai:
@iluwatar I will take a look.

iluwatar:
It has been suggested that we take into use [Maven checkstyle](https://maven.apache.org/plugins/maven-checkstyle-plugin/) plugin.

npathai:
@iluwatar Yes good point. It will reduce PR review effort.

iluwatar:
Checkstyle was added in a033c079f7d1c49f1736fb9d87fc281deadc2dbd

iluwatar:
I updated the instructions at https://github.com/iluwatar/java-design-patterns/wiki/3.-Coding-conventions

npathai:
@iluwatar The style guide worked for me in Eclipse Mars. To make sure this is how I imported it:
`Java > Code Style > Formatter > Import`

iluwatar:
Are you importing this https://github.com/google/styleguide/blob/gh-pages/eclipse-java-google-style.xml? 

I'm getting "Problems reading profiles from XML Attribute name "data-pjax-transient" associated with an element type "meta" must be followed by the ' = ' character."

npathai:
Build no of eclipse: `20150621-1200`
I get a version compatibility warning while importing but once I skip it, works fine.

iluwatar:
Strange, I have the same version.

Jahhein:
I was able to import these settings into intellij ultimate edition using an eclipse preferences plugin and use the reformat whole project feature. I have yet to make sure I meet all standards with what I have added. Is this still an issue to be resolved?

iluwatar:
@Jahhein That is great news. Would you create a pull request?

npathai:
@Jahhein Yes its still unresolved. Awaiting your PR.

Jahhein:
Will do, double checking to ensure all is correct. Intellij has a problem with spacing after class headers, and brackets, even if it is in the settings.

mafagafogigante:
[...]

PR #241, opened 2015-09-05
Improved unit tests in dao module and added log4j dependency.

themoffster:


dao/src/main/java/com/iluwatar/dao/App.java
```diff
@@ -11,43 +13,38 @@
...
 	public static List<Customer> generateSampleCustomers() {
-		Customer customer1 = new Customer(1, "Adam", "Adamson");
-		Customer customer2 = new Customer(2, "Bob", "Bobson");
-		Customer customer3 = new Customer(3, "Carl", "Carlson");
-
-		List<Customer> customers = new ArrayList<Customer>();
+		final Customer customer1 = new Customer(1, "Adam", "Adamson");
+		final Customer customer2 = new Customer(2, "Bob", "Bobson");
+		final Customer customer3 = new Customer(3, "Carl", "Carlson");
+		final List<Customer> customers = new ArrayList<Customer>();
```
npathai: Nitpick: Verbosity can be reduced here by using `<>`.

dao/src/main/java/com/iluwatar/dao/Customer.java
```diff
@@ -37,34 +37,39 @@ public String getLastName() {
...
-        if (o == null || getClass() != o.getClass()) return false;
-
-        Customer customer = (Customer) o;
-
-        if (id != customer.id) return false;
-
-        return true;
+    public boolean equals(final Object o) {
+    	boolean isEqual = false;
+    	final Customer customer = (Customer) o;
```
npathai: Logic: We should not promote incorrect conventions. There should not be a blind cast to `Customer` in `equals()`.
```java
public boolean equals(Object o) {
  if (o == this) { 
    return true; 
  }
  if (!(o instanceof Customer)) {
    return false;
  }
  // Its safe to cast here. Then the usual equality checks
}

File: dao/src/main/java/com/iluwatar/dao/Customer.java [hunk 3]
```diff
@@ -37,34 +37,39 @@ public String getLastName() {
...
+        	isEqual = true;
+        }
+        return isEqual;
     }
-
+    
     @Override
     public int hashCode() {
-        int result = id;
+        int result = getId();
```
npathai: Question: Any specific reason to alter the `hashcode()` ?

dao/src/main/java/com/iluwatar/dao/CustomerDao.java
```diff
@@ -9,9 +9,9 @@
  */
 public interface CustomerDao {
 	
-    public List<Customer> getAllCustomers();
-    public Customer getCusterById(int id);
-    public void addCustomer(Customer customer);
-    public void updateCustomer(Customer customer);
-    public void deleteCustomer(Customer customer);
+    List<Customer> getAllCustomers();
+    Customer getCustomerById(final int id);
```
npathai: Suggestion: I don't think we need to document parameters to be `final`. It increases verbosity.

dao/src/main/java/com/iluwatar/dao/CustomerDaoImpl.java
```diff
@@ -27,31 +27,33 @@ public CustomerDaoImpl(List<Customer> customers) {
...
     }
 
 
     @Override
-    public void updateCustomer(Customer customer) {
-        if (customers.contains(customer)) {
-            customers.set(customers.indexOf(customer), customer);
+    public void updateCustomer(final Customer customer) {
+        if (getAllCustomers().contains(customer)) {
+        	final int index = getAllCustomers().indexOf(customer);
```
npathai: Style nit: Indentation

dao/src/test/java/com/iluwatar/dao/CustomerDaoImplTest.java
```diff
@@ -0,0 +1,81 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+
+import org.junit.Before;
+import org.junit.Test;
+
+public class CustomerDaoImplTest {
+
+	private CustomerDaoImpl impl;
```
npathai: Shouldn't we call `CustomerDaoImpl` as `InMemoryCustomerDao`. It represents what the class does.
@iluwatar Do you think we should do that? We can add a `DatabaseCustomerDao` to it, if not now in future.

dao/src/test/java/com/iluwatar/dao/CustomerDaoImplTest.java
```diff
@@ -0,0 +1,81 @@
...
+	@Before
+	public void setUp() {
+		customers = new ArrayList<Customer>();
+		customers.add(CUSTOMER);
+		impl = new CustomerDaoImpl(customers);
+	}
+	
+	@Test
+	public void deleteExistingCustomer() {
+		assertEquals(1, impl.getAllCustomers().size());
```
npathai: Suggestion: If we are improving tests why not go a step further and add junit-hamcrest dependency as well. It will make this assertion read like
`assertThat(impl.getAllCustomers().size(), equals(1));`

dao/src/test/java/com/iluwatar/dao/CustomerDaoImplTest.java
```diff
@@ -0,0 +1,81 @@
...
+		impl.deleteCustomer(CUSTOMER);
+		assertTrue(impl.getAllCustomers().isEmpty());
+	}
+	
+	@Test
+	public void deleteNonExistingCustomer() {
+		final Customer nonExistingCustomer = new Customer(2, "Robert", "Englund");
```
npathai: Suggestion: 
```java
int initialCustomerCount = impl.getAllCustomers().size();
Customer nonExistingCustomer = new Customer(2, "Robert", "Englund");
impl.deleteCustomer(nonExistingCustomer);

assertThat(impl.getAllCustomers().size(), equals(initialCustomerCount));
```
This documents the intention more clearly IMO.

dao/src/test/java/com/iluwatar/dao/CustomerDaoImplTest.java
```diff
@@ -0,0 +1,81 @@
...
+		final Customer customer = new Customer(CUSTOMER.getId(), newFirstname, newLastname);
+		impl.updateCustomer(customer);
+		final Customer cust = impl.getCustomerById(CUSTOMER.getId());
+		assertEquals(newFirstname, cust.getFirstName());
+		assertEquals(newLastname, cust.getLastName());
+	}
+	
+	@Test
+	public void updateNonExistingCustomer() {
+		final int nonExistingId = 999;
```
npathai: Logic: `final int nonExistingId = getNonExistingCustomerId();`

dao/src/test/java/com/iluwatar/dao/CustomerDaoImplTest.java
```diff
@@ -0,0 +1,81 @@
...
+		impl.updateCustomer(customer);
+		assertNull(impl.getCustomerById(nonExistingId));
+		final Customer existingCustomer = impl.getCustomerById(CUSTOMER.getId());
+		assertEquals(CUSTOMER.getFirstName(), existingCustomer.getFirstName());
+		assertEquals(CUSTOMER.getLastName(), existingCustomer.getLastName());
+	}
+	
+	@Test
+	public void addCustomer() {
+		final Customer newCustomer = new Customer(3, "George", "Patton");
```
npathai: Missed case, what happens if we add an already existing customer? Logically it should fail. But our implementation does not. It will just add it to list.

dao/src/test/java/com/iluwatar/dao/CustomerTest.java
```diff
@@ -0,0 +1,87 @@
...
+	private static final String FIRSTNAME = "Winston";
+	private static final String LASTNAME = "Churchill";
+	
+	@Before
+	public void setUp() {
+		customer = new Customer(ID, FIRSTNAME, LASTNAME);
+	}
+	
+	@Test
+	public void getIndex() {
```
npathai: Suggestion: Personally I don't see any value in testing plain getters and setters. It feels like forced tests. All we can test in this class is `equals()` and `hashcode()` contract, because that's important. Do you agree @iluwatar ?

dao/src/test/java/com/iluwatar/dao/CustomerDaoImplTest.java
iluwatar: :+1:

dao/src/test/java/com/iluwatar/dao/CustomerTest.java
iluwatar: Yes, I agree. Testing setters and getters gives marginally better test coverage but increases code mass and maintenance effort. Overall the effect is negative.

npathai:
Ping @themoffster

themoffster:
Feel free to reject the changes, it makes no difference to me. 
The arguments I would make though is that there is absolutely no unit test coverage currently - if you have better alternatives feel free to use them instead.

The caveat to that is the .equals() and .hash() methods which I agree are incorrect and those changes should be rejected.

Testing getters and setters is valid IMO. I see the argument of not wanting them in classes with a lot of class variables but this is not the case here.

Finally, fields should be declared as final as that is best practice as you can see in reports from Sonarqube - with this project being a sample reference for other users, these best practices should be incorporated.

Cheers

npathai:
@themoffster We are not rejecting your changes. We appreciate that you are putting effort in contributing better tests. As far as testing getter and setter goes, it comes to personal taste. All the other points made are valid IMO and we await your responses on them, so that we can improve further. Even we want to be better FYI #224 .

themoffster:
closing this pull request to make changes based on feedback

Issue #197, opened 2015-08-10
Accepting presentations for the design patterns


zafarella:
@iluwatar What do u think if we create presentations for these nice collections of design patterns.
Either one for each for one for all. Something based on impress.js or similar would be great I think.
Happy to contribute.
Kiitos

npathai:
@zafarella Presentations are nice. I think we can also have Refcardz similar to [Javazone](https://dzone.com/refcardz/design-patterns). But I think #56 can get closer to Refcardz.
As far as presentation goes you can prepare first draft with 2-3 patterns and share with us, so that we can have a discussion on it.

iluwatar:
I agree we need better presentation. There is something boiling in #56 but I'm not sure what is the status of that. @zafarella I agree with @npathai Feel free to demostrate your ideas.

zafarella:
Yep, thanks, whenever will find some time - will implement one presentation. The idea behind is to simply open the repository and start sessions about the patterns. It would be nice I think when you have some classes to do or simply want to go through some patterns in a wide screen.

npathai:
:+1:

iluwatar:
Sounds awesome. I'm looking forward to see the first presentation.

npathai:
Ping @zafarella Any luck?

zafarella:
Today :). Singleton on it's way.

npathai:
Wow we are so excited to see the first presentation :+1:

iluwatar:
I saw the first presentation at https://github.com/iluwatar/java-design-patterns/pull/244 and I love the idea :+1:

npathai:
:+1: from me as well. It is impressive for first draft and we can surely take it forward and make it even better. I have some concerns regarding the snaps and material being out of sync from our codebase as our code evolves.

iluwatar:
Yes, I share your concern.

npathai:
I was thinking can't we enhance the website in future to be usable in presentation form. We can have something like presentation mode with pattern listings. That will keep content up to date. I don't know if it is feasible though.

noobxgockel:
@npathai as i wrote in PR #244 allready:
> Hmmm... We could automate the process of generating the presentation with the data visible on the website. I highly doubt it would be a good presentation though...

[...]

PR #244, opened 2015-09-07
add draft presentation for the singleton #197

zafarella:
As per discussion at #197


zafarella:
note ready yet. Pls wait a bit to finish it.

zafarella:
Take a look guys, how will you like it. Happy to address any issues/dislikes/mistakes.
Thanks

iluwatar:
What I did was go to the presentation directory and start Python web server

`python -m SimpleHTTPServer`

Then open browser and go to `http://localhost:8000`

iluwatar:
It would be nice to reuse portions of the data that we already have in `index.md`.

noobxgockel:
Hmmm... We could automate the process of generating the presentation with the data visible on the website. I highly doubt it would be a good presentation though... We (or more like mr. penguin here) could create presentations manually. The only problem with that would be that is allways out of date, because our repo is changing/adding patterns rapidly.

Dont let yourself be held down @zafarella. But in my honest opinion it wouldnt have much of a use. I only recall learning about patterns in a presentation while i was learning scala (monads, monadics). Presentations are, for me atleast, to be presented by a real person while clicking through the slides.

But if you want to, i can help you integrate your Presentations into the website C:

zafarella:
@noobxgockel thanks for the comments. This is exactly for the making presentation by a person. Sometimes you don't have time to prepare a presentation, but the material already exists, you just need to deliver it to audience. This is why I thought about having in this collection presentation, so you can just open it and start the class.

In regards the outdates - I think we can maintain it, there is no better option so far that I can think of.

@iluwatar you don't really need to run a server - just open the html file in browser. I will reuse the content of index.md.

I think would be better to move presentation and the file into etc directory. Will do that. Also what do you think if there will be umbrella (presentation) file with all patterns somewhere in the root folder.

iluwatar:
Let us know when you are ready for review.

zafarella:
I will try to fix reading the code into HTML page from file(s) and it is ready to be reviewed. Although I am not an HTML expert)

iluwatar:
There has been no progress during the last 30 days, so I will close this PR as expired.

PR #246, opened 2015-09-09
Add checkstyle into maven

zafarella:
This will add checkstyle plugin. It will not fail the build, but print warnings.
Align with styles could be done in separate branch.


noobxgockel:
In case you didnt know:
https://google.github.io/styleguide/javaguide.html#s2.3.1-whitespace-characters
TL;DR Java devs hate Tabs, dont ask me why...
Thats also why travis fails.

zafarella:
travis failed because log is too long and it assumed that this is due to test failure.
Local `mvn validate` successes but bunch of warnings, coz code is not styled according rules.

zafarella:
Should I do you think reformat java files according to styling guides?

noobxgockel:
Ohh now i get it... didnt think travis would fail just because the file is too large.
Well you couldnt possible format all the code, thats a huge task, but i'd suggest muting all the warnings and only printing the error (if thats possible with the maven checkstyle plugin).

We dont really need the warnings when travis is building the thing right?
I hope im right :S

iluwatar:
Please catch up the branch with origin/master and I will review. Is there an issue for this?

zafarella:
Synced with master. Please note that checkstyle is "off" and will not report anything. Each java file need to be aligned according to the rules of checkstyle. It can be turned on by switching flag to `<consoleOutput>false</consoleOutput>` to `true`.

npathai:
Yes once we format all code we will set the flag

iluwatar:
Thanks @zafarella

Issue #247, opened 2015-09-10
add getAttackPower method to pattern decorator


liangpengfei:
Add getAttackPower method to pattern decorator may make it more easily to understand .

when the troll not add smart decorator , his attack power is 10 ,but when we build the smartTroll , his attack power is 10(the normal troll's attack power) + 20(because smart). So ,it's more easy to understand .

PR #253, opened 2015-09-23
#247 adding getAttackPower method to pattern decorator

mgalushka:
As proposed on original description


iluwatar:
Thank you for the contribution!

PR #254, opened 2015-09-24
Refactor singleton according to checkstyle rules

zafarella:
This will eliminate all checkstyle warning according to checkstyle rules defined. Applicable just to singleton.


iluwatar:
Thank you!

PR #269, opened 2015-10-19
Add unit test to callback method

iamrichardjones:
add a unit test to callback pattern. Assert that the callback method is called using a counter


iluwatar:
Thank you :+1:

PR #275, opened 2015-10-28
Fix indentation

eug:


iluwatar:
This doesn't fix anything. We're using Google coding conventions as described here https://github.com/iluwatar/java-design-patterns/wiki/03.-Coding-conventions

PR #287, opened 2015-11-13
log4j.xml relocated and excluded from JAR

l-cortese:
`log4j.xml` relocated due to visibility issues and excluded from generated JAR. `pom.xml` formatted according to coding conventions


iluwatar:
Hi @l-cortese Can you explain a bit how it works?

dao/pom.xml
```diff
@@ -1,24 +1,69 @@
...
+      <groupId>log4j</groupId>
+      <artifactId>log4j</artifactId>
+    </dependency>
+  </dependencies>
+
+  <build>
+
+    <!-- 
+    log4j.xml file will be copied both in ${project.build.outputDirectory} 
+    and ${project.build.outputDirectory}. Thanks to Sean Patrick Floyd
```
iluwatar: Twice the same?

l-cortese:
Ok, there was a typo in the comment at line 29 of pom.xml, it should be 

> log4j.xml file will be copied both in ${project.build.outputDirectory} and ${project.build.**directory**}.

`${project.build.outputDirectory}` is `target/classes`
`${project.build.directory}` is `target`

`log4j.xml` must be copied under `target/classes` for the project to run correctly in Eclipse (nothing new here). `jar:jar` plugin, by default, makes up a JAR archive by copying the tree structure under `target/classes`, so it would include `log4j.xml` too. To **exclude** it from the JAR, I configured the plugin in the pom and "moved" `log4j.xml` to same level dir `target`, so it can be used (and modified if needed) without affecting the JAR.

l-cortese:
Now... if I wanted to correct the comments, should I treat it like a totally new PR?

l-cortese:
ok, just pushed corrections to the master branch in my repo

iluwatar:
Great, thanks for the explanation. So now that the config file is outside the jar it is possible to modify it and run the `dao.jar` with the altered configuration with some command?

l-cortese:
Exactly, you can run `dao-1.8.0-SNAPSHOT.jar` via commandline specifing some options:

    java -Dlog4j.configuration=file:.\log4j.xml -cp .\dao-1.8.0-SNAPSHOT.jar com.iluwatar.dao.App

paying attention to include referenced libraries in the classpath

iluwatar:
Thank you. It works as you explained.

Issue #63, opened 2015-06-23
Twin design pattern


iluwatar:
http://www.ssw.uni-linz.ac.at/Research/Papers/Moe99/Paper.pdf

hoswey:
This pattern is very interesting, please let me work on it.

iluwatar:
@hoswey go ahead, it's all yours.

PR #289, opened 2015-11-14
implements Twin design pattern #63

hoswey:
please help review


hoswey:
The coverage checking is fail even when I make some trivial changed, T_T.

Issue #286, opened 2015-11-11
No main method in the Abstract Factory example


iluwatar:
There is no `main` method in the Abstract Factory example.

tainarareis:
I'm working on it!

jwerba:
@tainarareis do you want me to help you?

PR #291, opened 2015-11-14
Issue #286: Added main method in Abstract Factory Example

tainarareis:
I think it would be better to rename the App class to something like: Factory Producer. And the main method should be in the real App file as unique method. What do you think?


iluwatar:
I agree with your refactoring proposal. I will accept if you make another PR for that.

iluwatar:
This looks good. Thank you for the help!

tainarareis:
Thank you for accepting! Witch tool do you use for making class diagrams? Is it free?

iluwatar:
It is described in the developer wiki, here https://github.com/iluwatar/java-design-patterns/wiki/01.-How-to-contribute

PR #308, opened 2015-11-30
Annotation Config was added.

neonds:
Hi, every one. Hoswey I saw that you use Spring Data. Maybe is a good idea to use a Spring annotation based config too.

https://github.com/hoswey/java-design-patterns/pull/1/files

Please check


iluwatar:
Thanks @neonds for the annotation configuration example

PR #312, opened 2015-12-06
Add proper unit tests for chain, bridge and builder pattern #293

fluxw42:
Added proper unit tests for chain, bridge and builder pattern #293

I did some minor modifications to the chain implementation, since there was nothing much to test. Request now has a method to check if someone in the chain handled it.


iluwatar:
Great job @fluxw42 :+1:

PR #313, opened 2015-12-07
Update index.md

jptiancai:
add source code link of `java.lang.Runtime#getRuntime() `


npathai:
I have doubt whether the link adds value.

iluwatar:
I agree with @npathai I'm sorry but this has to be rejected.

jptiancai:
well, if add a  link of doing searching code is convenient.

PR #315, opened 2015-12-11
Just  Just a quick fix for the Front-controller pattern.

Deses:
Hiya!

I was just checking out this pattern and it returned three times the same string 
> Error 500
> Error 500
> Error 500

Correct output should be
>Displaying archers
>Displaying catapults
>Error 500

The problem was the name of the package was not the correct one in FrontController.java, line 28.

The class should be probably be uptaded to not get the Class names from hard-coded Strings.


iluwatar:
Thank you for the fix @Deses

PR #316, opened 2015-12-11
Added more JUnit tests

fluxw42:
Add proper unit tests for composite, dependency-injection, double-checked-locking, double-dispatch and facade. #293


iluwatar:
Top quality contribution! Thanks @fluxw42

PR #320, opened 2015-12-21
New unit tests for seven additional patterns #293

fluxw42:
New unit tests for seven additional patterns #293 :
 - layers
 - lazy-loading
 - mediator
 - memento
 - model-view-controller
 - model-view-presenter
 - monostate


fluxw42:
Ouch, seems like there is a test failing, even though I ran every test locally. I'll fix it and create another pull request.

Issue #75, opened 2015-06-23
Readers-writer lock


iluwatar:
https://en.wikipedia.org/wiki/Readers%E2%80%93writer_lock

hoswey:
Please let me know if it is ok for me to work on this issue.

iluwatar:
Sure @hoswey go ahead :+1:

PR #326, opened 2015-12-26
implements the #75 reader writer lock

hoswey:
Please help to review,  it seems the test case is hard to write for such concurrent pattern. :(


iluwatar:
Review comments:

- The application does not exit correctly when `App.java` is run from IDE
- The example does not have proper unit tests
- The explanation of the pattern in `App.java` is very brief. Could the explanation be expanded a bit?

reader-writer-lock/index.md
```diff
@@ -0,0 +1,29 @@
+---
+layout: pattern
+title: Producer Consumer
```
iluwatar: The title is wrong

reader-writer-lock/src/main/java/com/iluwatar/reader/writer/lock/App.java
```diff
@@ -0,0 +1,82 @@
...
+   */
+  public static void main(String[] args) {
+
+    ExecutorService es = Executors.newFixedThreadPool(1000);
+    ReaderWriterLock lock = new ReaderWriterLock();
+
+    AtomicInteger index = new AtomicInteger(0);
+    IntStream.range(0, 100).forEach(i -> {
+      Runnable task = null;
+      if (ran.nextFloat() <= 0.6) {
```
iluwatar: Why use random numbers here? I think you can just allocate certain number of readers and writers.

iluwatar:
@hoswey you have my review remarks. Please comment on this thread once ready for another look.

reader-writer-lock/src/main/java/com/iluwatar/reader/writer/lock/ReaderWriterLock.java
```diff
@@ -0,0 +1,147 @@
...
+
+          if (isLockFree()) {
+            g.add(this);
+            break;
+          } else if (isWriterOwnThisLock()) {
+            waitUninterruptely(g);
+          } else if (isReaderOwnThisLock()) {
+            waitUninterruptely(g);
+          } else {
+            throw new RuntimeException("it should never reach here");
```
iluwatar: You should not throw raw `RuntimeException`s but subclasses of this. See [AvoidThrowingRawExceptionTypes](http://pmd.sourceforge.net/pmd-4.3.0/rules/strictexception.html)

reader-writer-lock/src/main/java/com/iluwatar/reader/writer/lock/ReaderWriterLock.java
```diff
@@ -0,0 +1,147 @@
...
+    public void unlock() {
+
+      synchronized (g) {
+        g.remove(this);
+        g.notifyAll();
+      }
+    }
+  }
+
+  private boolean isWriterOwnThisLock() {
```
npathai: Better naming convention would be `doesWriterOwnThisLock()` and `doesReaderOwnThisLock()`

reader-writer-lock/src/main/java/com/iluwatar/reader/writer/lock/ReaderWriterLock.java
```diff
@@ -0,0 +1,147 @@
...
+    return g.contains(readLock);
+  }
+
+  private boolean isLockFree() {
+    return g.isEmpty();
+  }
+
+
+
+  private void waitUninterruptely(Object o) {
```
npathai: Typo: `waitUninterruptibly()`

reader-writer-lock/src/main/java/com/iluwatar/reader/writer/lock/ReaderWriterLock.java
```diff
@@ -0,0 +1,147 @@
...
+
+/**
+ * Class responsible for control the access for reader or writer
+ */
+public class ReaderWriterLock {
+
+  /**
+   * Mutex for reader
+   */
+  private Object r = new Object();
```
npathai: single character names should not be promoted, I would propose `mutex` or `readMutex` something in that line which also describes what it is about. And then the javadoc on private variable will not be needed.

reader-writer-lock/src/main/java/com/iluwatar/reader/writer/lock/ReaderWriterLock.java
```diff
@@ -0,0 +1,147 @@
+package com.iluwatar.reader.writer.lock;
+
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * Class responsible for control the access for reader or writer
```
npathai: Javadoc can be a bit more detailed here. "Allows multiple readers to hold the lock at same time, but if any writer holds the lock then readers wait. If reader holds the lock then writer waits." This lock is _not fair_.

reader-writer-lock/src/main/java/com/iluwatar/reader/writer/lock/ReaderWriterLock.java
```diff
@@ -0,0 +1,147 @@
...
+
+  /**
+   * Global mutex for reader or writer, use to save the holding object
+   */
+  private Set<Object> g = new HashSet<>();
+
+  /**
+   * Current reader count
+   */
+  private int readerCount = 0;
```
npathai: rename the private variable to `currentReaderCount` and remove the javadoc comment.

reader-writer-lock/src/main/java/com/iluwatar/reader/writer/lock/ReaderWriterLock.java
npathai: A better approach is to throw `AssertionError`. It should be used in such cases.

hoswey:
Thanks @iluwatar  and @npathai 's comments, the fix is done, please help to review. Thanks.

iluwatar:
Great work with this pattern and the tests. Thanks @hoswey once again :+1:

iluwatar:
@hoswey it seems that the unit tests are failing on Travis builds. Can you please check issue https://github.com/iluwatar/java-design-patterns/issues/343

npathai:
Good work @hoswey . I will have to go through the logic once more as you already stated that concurrency patterns are quite complex and there can be subtle issues lurking. If there is something I will let you know and create an issue for the same.

Issue #334, opened 2015-12-31
Callback implemented with lambdas


iluwatar:
In the Callback example show how to implement the same thing with Java 8 lambdas.

mikulucky:
I have added an example, not sure if this is really what you're asking for.

PR #337, opened 2016-01-01
Provide a simple example for Callback Pattern using Lambdas

mikulucky:
Created a Lambdas example for the callback pattern, using the same output of the traditional version. Have also added another unit test for issue #334


iluwatar:
Thank you @mikulucky

Issue #238, opened 2015-09-04
Using headings instead of bold text inside index.md


noobxgockel:
Currently we are using bold text as a heading of a chapter (Intent, Applicability, Real world examples, ...)
Example: ``` **Intent:** ```
Correct me if i'm wrong but i think this is a relic from the time before the great split. Nowadays we wont clutter the readme with too much headings.

I think turning those into Headings would improve the readability, the overview and, if i re-implement the TOC in the sidebar, even the navigation.

If we do that we should use at least the second level heading, as the first level is the name of the pattern.

noobxgockel:
@iluwatar, @npathai any thoughts on this ?
If this feature is wanted i could try and script something up.

iluwatar:
Good that you brought this up. When I look at it the bold text seems rather small for the purpose. :+1: for the bigger headings.

npathai:
:+1: from me too. Go for it.

fluxw42:
Just created [this script](https://gist.github.com/fluxw42/464820d08e3b567fda48) to update the bold titles with a level 2 header. Running this gives 3d642cd. Does that look ok to you?

noobxgockel:
With the PR merged and the wiki edited this issue is fixed.

iluwatar:
:+1: Looks great

PR #339, opened 2016-01-03
Use headings instead of bold text in index.md #238

fluxw42:
Just created [this script](https://gist.github.com/fluxw42/464820d08e3b567fda48) to update the bold titles with a level 2 header. Does that look ok to you? (Issue #238)

If this PR is accepted then the template on the [How-to-contribute](https://github.com/iluwatar/java-design-patterns/wiki/01.-How-to-contribute#new-pattern) wiki page should probably also be updated with the new level 2 heading.


noobxgockel:
This is awesome! :smiley: 
Thank you very much, i seem to have totally forgot about that issue.
The other 2 seem to have allready accepted the change in the referenced issue so i will just merge it and quickly change adjust the wiki template.

PR #345, opened 2016-01-14
Fix the wrong uml .png name in decorator

azhuzhu:
change decorator_1.png to decorator.png


iluwatar:
Thanks @azhuzhu for spotting this!

npathai:
Thanks for the fix @azhuzhu

Issue #358, opened 2016-01-27
Add MIT license header to source files


iluwatar:
One option: https://github.com/mycila/license-maven-plugin

mikulucky:
Pull request opened, ready for review when you get a chance. :smile:

PR #360, opened 2016-01-27
Pull Request for adding License Plugin and Generate #358

mikulucky:
Currently have the plugin set to generate the license on **install** however this is easy to change. This ensures that hopefully this will never be a review comment for a missing license. For issue #358


iluwatar:
Tested this a bit. It seems to work as expected. I think running `license:format` in install phase is a sensible default. Thank you @mikulucky :+1:

PR #386, opened 2016-02-18
Dao Pattern : Create object with reference to Interface

amitbhoraniya:
In Dao pattern DaoImpl object is created with reference to dao interface.


iluwatar:
Thanks @amitbhoraniya for spotting this

Issue #351, opened 2016-01-16
Plugin pattern


iluwatar:
http://martinfowler.com/eaaCatalog/plugin.html

skylermj-252:
Hello I am ready to tackle this issue. Is there any information I need to read to understand this issue?

npathai:
This pattern is described in Patterns of Enterprise Application Architecture book.

Brief description about the pattern:
 - Separated Interface: Some behaviors implementation must be chosen dynamically based on runtime environment. Different environments include Unit Testing, Stress Testing, IoT, Production. Ex of such a requirement is ID generator, in unit testing environment in memory counter based solution should be chosen and in production DB based counter must be chosen.
 - Dynamic linking: A plugin implementation is not known during compile time. The actual implementation is chosen based on configuration at runtime, and the class is loaded using reflection. The configuration can be a text file containing full path of plugin class.
- Reconfiguration: Reconfiguration will not require rebuild, as we have them loaded dynamically. I am thinking about a valid use case for this. I will comment as soon as I find an appropriate use case.

npathai:
Random thoughts on reconfiguration: Java does not allow reloading same class. So I guess reconfiguration means switching from one implementation to other. Like switching from In Memory Counter implementation to DB based counter.

geniusgeek:
The best implementation approach for this pattern is to use annotations,
reflection and dynamic binding in java, and Abstract Factory combined with
Strategy pattern so as to give the desired flexibility.
Although i am really busy presently, but i will love to render my
assistance in any way.

Cheers.

iluwatar:
This issue is free for taking again.

iluwatar:
Here are some guidelines http://stackoverflow.com/questions/465099/best-way-to-build-a-plugin-system-with-java

iluwatar:
@abhi-agr you can write the implementation from scratch or utilize @npathai's solution if you wish. Will you work on this?

abhi-agr:
I am planning to work on this. But I think I need to compare different strategies - OSGI and JPF to find out the better implementation.

Let me come up with a design and later I can move to implementation

iluwatar:
Ok @abhi-agr

iluwatar:
Issue free again

jasciiz:
please check it.

iluwatar:
@jasciiz are you still working on this?

lkpengcs:
hi, can I work on this issue?

[...]

PR #395, opened 2016-03-02
Added Plugin Pattern Example Source

amitjoy:
Solves #351


iluwatar:
Review remarks:
- you should add plugin as a module to the parent `pom.xml`

plugin/index.md
```diff
@@ -0,0 +1,21 @@
...
+categories: Enterprise
+tags:
+ - Java
+ - Difficulty-Intermediate
+ - Extension Points
+ - Enterprise Architecture Pattern
+---
+
+## Intent
+Ensure configuration update in a dynamic way and provide different configurations for different environments
```
iluwatar: Mr. Fowler's description: "Links classes during configuration rather than compilation"

plugin/src/main/java/com/iluwatar/plugin/Client.java
```diff
@@ -0,0 +1,44 @@
...
+ */
+package com.iluwatar.plugin;
+
+import com.google.inject.Guice;
+import com.google.inject.Injector;
+import com.iluwatar.plugin.config.modules.ProductionModule;
+
+/**
+ * Client Application
+ */
```
iluwatar: Please add pattern description and implementation notes here. As in the other examples.

iluwatar:
@amitjoy you have my review comments. Please comment on this thread once you've addressed them.

npathai:
Pattern described needs to have the ability to dynamically reload the plugin which can change the implementation I guess that is where the power lies. Presently it is just about programming to an interface, choosing the right implementation and sticking to it.

@amitjoy Also I am not sure whether such a small project needs guice dependency. The lower the use of other frameworks the better it is to understand the classical pattern. @iluwatar What do you think about that?

iluwatar:
I agree with @npathai this example could be done without additional frameworks such as Guice

amitjoy:
According to the implementation I have used, I am using **Dependency Inversion** (more precisely **Dependency Injection**) principle to manage dependencies at runtime by providing different configurations. Dependency inversion itself makes use of Program to an Interface principle for injecting dependencies dynamically. Under the hood, the actual object instances are created and injected using Java Reflections.
Apart from it, the use-case as mentioned by Martin Fowler, is an example usage of Testing Pattern to mock behaviours for System Under Test (SUT). You can find many examples to use Google Guice in such scenarios. I understand your need for simplified source (without Google Guice) of this Plugin Pattern but I believe people can also understand the use of DI Framework in implementing Enterprise Patterns.

npathai:
@amitjoy I understand your use of guice. But the scenarios that you explained can be achieved using manual constructor injection. Or we need setter injection here coz we change implementation at runtime. Using guice would be more appropriate if example demanded a DI container.

iluwatar:
Please tell us when you're ready for another review.

iluwatar:
@amitjoy this pull request is about to expire due to inactivity. Please comment if you still work on this.

amitjoy:
Hi @iluwatar,

I was a bit busy with some other stuffs. I would make the necessary changes to this pull request.

iluwatar:
Ping @amitjoy :neckbeard:

iluwatar:
@amitjoy I'm about to close this pull request due to no activity during previous month.

iluwatar:
Closed due to inactivity.

Issue #332, opened 2015-12-31
Factory Method implemented with lambdas


iluwatar:
In the Factory Method example show how to implement the same thing with Java 8 lambdas.

raviteja1992y:
I started working on this.

npathai:
We have decided to keep this pattern classical and not use lambdas. So closing it.

PR #396, opened 2016-03-05
Factory Method implemented with lambdas #332

raviteja1992y:


raviteja1992y:
I have cleared the errors. 
Build and Tests were successful.

npathai:
@raviteja1992y Implementing factory method with Lambdas did not make much sense to me. I cannot see advantages of it.
@iluwatar Can we have a discussion on this before merging this PR?

geniusgeek:
I think lambdas have underlying implementation of factory method to ease the work of the programmers. Siting the work of Doug Schmidth , he clearly stated that the frequency of use of certain design patterns have come to be part of some languages. 
Therefore i believe that we shouldn't use Lambdas to implement factory design pattern

iluwatar:
I raised this issue a while ago and it seemed a good idea then to "modernize" this pattern to utilize lambdas. Now seeing this implemented I tend to agree that it may not be so wise idea after all. Keeping this design pattern "classic" seems to be the better option.

npathai:
Yes I kinda feel the same. And usually factories are not just about calling proper constructors but have the responsibility of wiring all the objects and dependencies together. Keeping it classic makes sense.

If we had implemented it then the factory interface could have default methods for returning factory implementations as lambda objects but that would violate OCP. Presently the App itself creates lambda expressions for creating objects of particular type, which is not the case in real applications. Usually actual factory to use depends on configuration and environment and different factories have different ways of wiring objects which App will have no idea of. I hope it makes sense.

raviteja1992y:
@npathai I agree. Factory method logic should be in the derived classes (OrcBlacksmith and ElfBlacksmith).

PR #398, opened 2016-03-08
Event driven architecture refactored.

legka:
1. Renamed Message to Event and Event to AbstractEvent
2. Generified Event and Handler to avoid casts in Handlers
3. Updated EventDispatcher to make unsafe configuration impossible (handler for wrong type)
4. Updated UML diagram accordingly


event-driven-architecture/src/main/java/com/iluwatar/eda/App.java
```diff
@@ -53,12 +53,12 @@
   public static void main(String[] args) {
 
     EventDispatcher dispatcher = new EventDispatcher();
-    dispatcher.registerChannel(UserCreatedEvent.class, new UserCreatedEventHandler());
```
legka: Here you could actually have something like this:
```
dispatcher.registerChannel(UserCreatedEvent.class, new UserCreatedEventHandler());
dispatcher.registerChannel(UserUpdatedEvent.class, new UserUpdatedEventHandler());
```

The updated dispatcher will not allow this.

event-driven-architecture/src/main/java/com/iluwatar/eda/framework/EventDispatcher.java
```diff
@@ -22,19 +22,16 @@
...
 
 /**
  * Handles the routing of {@link Event} messages to associated handlers.
  * A {@link HashMap} is used to store the association between events and their respective handlers.
- *
  */
 public class EventDispatcher {
 
-  private Map<Class<? extends Event>, Handler<?>> handlers;
+  private Map<Class<? extends Event>, Handler<? extends Event>> handlers;
```
legka: This map declaration and the `registerHandler()` signature below, both ensure only 'type-safe' combination of event type and handler can be registered.

npathai:
:+1: Good changes. Thanks for your contribution.

legka:
My pleasure )

PR #405, opened 2016-03-18
Data transfer object

raviteja1992y:
Hi,
I worked on this pattern. Can some one review this and share your view?


npathai:
Initial reviews:
 - Javadocs for some classes are missing. Not all *Bo classes have documentation
 - There are no test cases.
 - Class diagram is non standard
 - Will need discussion whether we should include data diagram. Look at https://github.com/iluwatar/java-design-patterns/wiki/01.-How-to-contribute for more on this.
 - Main class should be named App.java for following same standard
 - index.md needs to be a bit more descriptive. User should be able to understand the usecase
 - Add json library dependency in parent pom and use it in child pom.

I have not yet gone through the logic, I will review it at latest.

npathai:
@raviteja1992y You have my initial reviews. Let me know when you have addressed the changes. Or in case there are any questions or doubts.

geniusgeek:
I understand that you are trying to implement an elegant DTO, instead of
using a Json Array and Json Parser, you would have made use of Java
Reflection and Jackason Library to create the instance of the object and
set the field data. Otherwise you would have kept it simple as implemented
in tutorials point.


Cheers

raviteja1992y:
@npathai I have made the changes you mentioned. Can you have a look at them once? I have a few questions.
- What do you mean by *Bo classes in the first point?
- Why dont you want to have the data diagram?

raviteja1992y:
@geniusgeek I am not trying to implement elagant DTO.

I think that the tutorialspoint example is more confusing. They added lot of unnecessary logic in Business object like data insertion in constructor, update data etc.. I think it is deviating from explaining the data transfer object.

Thats why I used the BO only to fill the DTO and kept the data seperate.

npathai:
@raviteja1992y I meant the Business Object classes don't have javadocs. And its not that I don't want a data diagram, we can place it below class diagram.

iluwatar:
@raviteja1992y @npathai what is the status of this pull request? I see there are conflicts that should be resolved. Is it pending for review or fixes?

raviteja1992y:
@iluwatar @npathai Initial review is done and i have made changes accordingly. 
I solved the merge conflicts. But then came some problem with build ( Could not find resource 'checkstyle-suppressions.xml' ). I will fix it.

raviteja1992y:
@npathai Hi. I have fixed the issues except the use cases(haven't found many). Can you review it?

data-transfer-object/src/test/java/com/iluwatar/data/transfer/object/DtoTest.java
```diff
@@ -0,0 +1,54 @@
...
+  
+  /*
+  * Printing all orders
+  */
+  @Test
+  public void testAllOrders() {
+    OrderBo bo = new OrderBo();
+    OrderDto dto0 = bo.getOrderDetails(0);
+    OrderDto dto1 = bo.getOrderDetails(1);
+    OrderDto dto2 = bo.getOrderDetails(2);
```
npathai: There are no assertions in the test. I would suggest you to assert on the result. You can assert on the details of the order.

data-transfer-object/src/main/java/com/iluwatar/data/transfer/object/OrderBo.java
```diff
@@ -0,0 +1,95 @@
...
+   */
+  private void readData() {
+    JSONParser parser = new JSONParser();
+    try {
+      this.orders = (JSONArray) parser.parse(new FileReader("data/orders.json"));
+      this.customers = (JSONArray) parser.parse(new FileReader("data/customers.json"));
+      this.products = (JSONArray) parser.parse(new FileReader("data/products.json"));
+    } catch (IOException e) {
+      System.out.println(e.getMessage());
+      System.exit(-1);
```
npathai: I would suggest wrapping this up in a `RuntimeException` and throwing it, rather than doing `System.exit(-1)`. That would look more elegant.

data-transfer-object/src/main/java/com/iluwatar/data/transfer/object/OrderDto.java
```diff
@@ -0,0 +1,105 @@
...
+ * Products, customers and orders data are saved in their respective json files. 
+ * <p>
+ * products.json is an array of product object. Product object contains product name, price, etc..
+ * <p>
+ * customers.json is an array of customer object. Customer object contains customer name, address, etc..
+ * <p>
+ * Orders.json is an array of order object. Order contains the unique identifier of the customer who 
+ * placed the order and unique identifier of the product he/she ordered.
+ * <p>
+ * I considered the index of product in products json array as its unique identifier for simplicity in coding.
```
npathai: > I considered the index of product in products json array as its unique identifier for simplicity in coding.

Can be rephrased as "For simplicity the indexes of Customers, Products and Orders are considered to be their unique identifiers."

data-transfer-object/src/main/java/com/iluwatar/data/transfer/object/OrderDto.java
```diff
@@ -0,0 +1,105 @@
...
+ * customers.json is an array of customer object. Customer object contains customer name, address, etc..
+ * <p>
+ * Orders.json is an array of order object. Order contains the unique identifier of the customer who 
+ * placed the order and unique identifier of the product he/she ordered.
+ * <p>
+ * I considered the index of product in products json array as its unique identifier for simplicity in coding. 
+ * Same is for the customer in customers json array and order in orders json array.
+ * <p>
+ * OrderDto is the Data transfer object for order details i.e orderIndex, productName, customerName etc..
+ * CLient requests for the OrderDto object of a specific orderId through OrderBo object.
```
npathai: Typo: CLient.
It should be Client.

data-transfer-object/src/main/java/com/iluwatar/data/transfer/object/OrderBo.java
```diff
@@ -0,0 +1,95 @@
...
+
+import java.io.FileReader;
+import java.io.IOException;
+import org.json.simple.JSONArray;
+import org.json.simple.JSONObject;
+import org.json.simple.parser.JSONParser;
+
+/**
+ * Client requests the data transfer object through Business object.
+ * Business object fills the data transfer object.
```
npathai: In our demonstration of DTO we have direct dependency on JSON files. I wonder if we should atleast create Service Layer abstraction OrderService so that we can decouple OrderBo. For now we can have one implementation of OrderService called `FileOrderService` which reads data from a file and then we can also add an actual remote service call. Let me know what you think @raviteja1992y

data-transfer-object/index.md
```diff
@@ -0,0 +1,32 @@
...
+folder: data-transfer-object
+permalink: /patterns/data-transfer-object/
+categories: Creational
+tags: 
+ - Java
+ - Difficulty-Beginner 
+---
+
+## Also known as
+Virtual Constructor
```
npathai: I doubt whether it is called Virtual constructor. Virtual constructor is a term for Factory pattern. DTOs are also known as Transfer Objects or vaguely as Value Objects. Can we correct this?

data-transfer-object/index.md
```diff
@@ -0,0 +1,32 @@
...
+![alt text](./etc/data-diagram.png "data transfer object")
+
+## Applicability
+Use this pattern when
+
+* you need to make multiple calls to remote, you can reduce the number of calls by tranferring all the data at once as a data transfer object.
+
+## Credits
+
+* https://msdn.microsoft.com/en-us/library/ff649585.aspx
```
npathai: https://msdn.microsoft.com/en-us/magazine/ee236638.aspx

We should list down the pros and cons of the pattern as well.

data-transfer-object/index.md
```diff
@@ -0,0 +1,32 @@
...
+## Class UML
+![alt text](./etc/dto_uml.png "data transfer object")
+
+## Data structure
+![alt text](./etc/data-diagram.png "data transfer object")
+
+## Applicability
+Use this pattern when
+
+* you need to make multiple calls to remote, you can reduce the number of calls by tranferring all the data at once as a data transfer object.
```
npathai: It is not 100% true that we need DTOs when there are multiple calls. There are domain objects in the service layer and those domain objects if serializable can be directly passed on to the client. So that would also eliminate the need for multiple remote calls. 

The are two situations where DTOs are useful, first is when the presentation layer needs a widely different view than the domain service layer. And if the domain layer details should be hidden from the client of the service then also DTOs are useful. 

This info is useful because DTOs are highly misused in J2EE context.

npathai:
@raviteja1992y You have my review comments. Can we please resolve this PR soon?

raviteja1992y:
@npathai . Thank you for the review. Sorry for responding late. I have some other work. I will get back to this after one/two weeks.

npathai:
@raviteja1992y Any updates? We will have to close this PR if it remains inactive.

npathai:
I am closing this PR due to inactivity. Create a new PR after you make the changes suggested.

Issue #404, opened 2016-03-16
Refactor DAO pattern


npathai:
I have created a branch to refactor DAO pattern https://github.com/iluwatar/java-design-patterns/tree/refactor-dao

Done
 - Refined interface according to Java 8
 - Feedback from CUD operations
 - Renamed existing implementation to `InMemoryCustomerDao`

To be Done
 - Add DBDao (But I was wondering who should create the schema)

Does App needs to have knowledge about the datasource that will be utilized so that it can create schema? Does anyone have hands on experience on this?

PR #407, opened 2016-03-23
Solves #404 Refactor dao

npathai:
List of changes:
 - Two implementations of `CustomerDao` interface. `DbCustomerDao` and `InMemoryCustomerDao`.
 - `CustomerDaoImpl` is renamed to `InMemoryCustomerDao`.
 - Use of Hierarchical Context Runner for making tests more readable.
 - Updated class diagram.
 - Schema is created by App before using `DbCustomerDao`.


npathai:
@iluwatar Please go through the changes. I have changed test cases a bit to use my style of unit testing. The method names are verbose and communicate the behavior under test.

dao/pom.xml
```diff
@@ -44,6 +44,19 @@
       <groupId>log4j</groupId>
       <artifactId>log4j</artifactId>
     </dependency>
+    <dependency>
+        <groupId>com.h2database</groupId>
+        <artifactId>h2</artifactId>
+        <scope>compile</scope>
```
iluwatar: `compile` is the default scope so no need to define it

iluwatar:
Review remarks:
- The SQL for creating/dropping schema is duplicated multiple times here and there. Would it be possible to put this in one place e.g. SQL-file and execute from there?
- Handling of `null` values. It's a matter of style and opinion but we could utilize `Optional` instead.

iluwatar:
@npathai you've got my review comments.

npathai:
@iluwatar 
 - Regarding SQL I will try to address it.
 - Regarding `Optional`, I gave it some thought while designing the interface according to java 8. But left it as is because Optional is useful when `null` can have two meanings. Such as `Map.get()`, where it can mean that key was not mapped to any value or key was mapped to a null value. There were a few links as well discussing it. I will attach it. Optional is not a replacement for null.

npathai:
@iluwatar Well after a lot of opinion reading I decided that for the sake of keeping interface fuctional ready, I will change the return type to be Optional.

But in reality I dont hate nulls. If documented correctly they are not a problem. Also creating an optional wrapper creates unnecessary GC pressure if the operation is heavily used.

npathai:
@iluwatar Done with the changes.

iluwatar:
@npathai these are great improvements :+1:

PR #412, opened 2016-03-26
Override annotation added

slawiko:
I think this is a typo

### DIFF
--- a/decorator/src/main/java/com/iluwatar/decorator/Troll.java
+++ b/decorator/src/main/java/com/iluwatar/decorator/Troll.java
@@ -29,6 +29,7 @@ package com.iluwatar.decorator;
  */
 public class Troll implements Hostile {
 
+  @Override
   public void attack() {
     System.out.println("The troll swings at you with a club!");
   }
@@ -38,6 +39,7 @@ public class Troll implements Hostile {
     return 10;
   }
 
+  @Override
   public void fleeBattle() {
     System.out.println("The troll shrieks in horror and runs away!");
   }


iluwatar:
In my opinion it is good to have the `@Override` annotation when implementing interface methods. This way we can utilize compiler checks to verify that the implementing method signature really matches the interface and is not a new method. We could enforce this with Checkstyle i.e. Checkstyle would give an error if interface method implementation would not specify `@Override`. @slawiko @npathai what do you think?

slawiko:
@iluwatar I thought it was already in your Checkstyle. Howbeit we MUST add it in checkstyle :)

iluwatar:
@slawiko it is not there yet, otherwise the build would not work. Maybe it is this one: http://checkstyle.sourceforge.net/config_annotation.html#MissingOverride Could you try adding it and see what happens?

slawiko:
@iluwatar yes of course, but not right now.

iluwatar:
No problem, just let us know then.

slawiko:
@iluwatar as I understood `<module name="MissingOverride"/>` works only if code contains JavaDoc tag `{@inheritDoc}`. But if I add this tag Imgur code inspection with Imgur still passes checkstyle: Imgur. I use Checkstyle for the first time, so I can be wrong. By the way, I tried without JavaDoc tag `{@inheritDoc}`, but result was identical. What do I do wrong?

iluwatar:
Looks like `MissingOverride` is not what we want. I'm not sure if Checkstyle has exactly what we need.

iluwatar:
My suspicion is verified in http://checkstyle.sourceforge.net/google_style.html

"6.1 @Override: always used" - "That validation could not be checked by Checkstyle. It's need to take a look as parent class. But Checkstyle have no way open or look at another Class file. "

iluwatar:
Thanks @slawiko for the improvement!

slawiko:
@iluwatar you're welcome!

Issue #34, opened 2015-03-21
Data Mapper Pattern


jesusvazquez:
Hi, I'm a new user to github and I would like to help to this project. I've been researching about a few patterns and the work you've done and I think it's so cool.

In my college I've developed an application which comunicates with a MySQL Database and I had to build a pattern to make that communication work. I'm talking about the Abstract Mapper.

I have an example here: https://github.com/jesusvazquez/java_examples/tree/master/AdvancedDatabaseSystems/src/modelo/DBMapper

I'm wondering if it would be a good idea to add this pattern to your project.

Best Regards.

fx773d:
Hi Vazques,

I'm a new user too and I think you should talk with the owner of the
project, I've just contributed to project. Sorry that I cannot help you.

Regards.

2015-03-21 16:14 GMT+02:00 Jesús Vázquez <notifications@github.com>:

> Hi, I'm a new user to github and I would like to help to this project.
> I've been researching about a few patterns and the work you've done and I
> think it's so cool.
>
> In my college I've developed an application which comunicates with a MySQL
> Database and I had to build a pattern to make that communication work. I'm
> talking about the Abstract Mapper.
>
> I have an example here:
> https://github.com/jesusvazquez/java_examples/tree/master/AdvancedDatabaseSystems/src/modelo/DBMapper
>
> I'm wondering if it would be a good idea to add this pattern to your
> project.
>
> Best Regards.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/iluwatar/java-design-patterns/issues/34>.
>



-- 
*Bener Çelenkoğlu*
*Departmant of Computer Engineering, Hacettepe University*
*Tel: (506)776-3457*

iluwatar:
@jesusvazquez Thank you for the interest in this project.

For the Abstract Mapper pattern you are proposing let's discuss about it on a bit higher level. Can you describe what is the intent of the pattern and when should we use it the same way as we have for the other patterns [here](https://github.com/iluwatar/java-design-patterns/blob/master/README.md)?

jesusvazquez:
Yes, 

This pattern allows you to map your database tables without using any framework, like Hibernate for example. If your application is small and you want to understand how the connections to the database work, this pattern helps you a lot because it's just an abstract class and a few Java programming concepts.

If you want to map a table you just have to fill 4 or 5 methods and then SELECT, INSERT, UPDATE or DELETE will work perfectly.

[...]

PR #417, opened 2016-04-06
Data Mapper #34

inbravo:
A basic data-mapper implementation based on discussion on #34


iluwatar:
Review comments:
- `StudentFirstMapper` and `StudentSecondMapper` seem very similar. Can we drop one of them to simplify the example? You wouldn't need to process command line parameters in `App` after the change.
- In `App` the conditional use of `Logger` is confusing. Could we just output the debug log without testing any conditions?
- `DataMapperException` seems to forward the call to `super` in most of its methods. In this case you can leave out the implementation because it's the default behavior.

data-mapper/src/main/java/com/iluwatar/datamapper/StudentDataMapper.java
```diff
@@ -0,0 +1,32 @@
...
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT
+ * NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+ * NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM,
+ * DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+package com.iluwatar.datamapper;
+
+import java.util.Optional;
+
```
iluwatar: Add JavaDoc

data-mapper/src/main/java/com/iluwatar/datamapper/DataMapperException.java
```diff
@@ -0,0 +1,79 @@
+/**
+ * The MIT License Copyright (c) 2016 Amit Dixit
```
iluwatar: Please keep the license header consistent

iluwatar:
@inbravo you have my review comments. Please notify me once you've addressed them.

inbravo:
- `StudentSecondDataMapper ` is dropped and `StudentFirsDataMapper ` is renamed to `StudentDataMapperImpl` for better readability.
-  Conditional usage of `logger` in App class is discarded.
-  `DataMapperException` is forwarding calls to `RuntimeException` which is a core Exception class but not any reference from implementation. This is helpful in detailing users about removing all data-mapper exception dependency from code. Please review again.
- Student equality test is added at `StudentTest`
- Your last point is not clear to me; please share more information on it.

I hope just a final review is required.

data-mapper/src/main/java/com/iluwatar/datamapper/DataMapperException.java
inbravo: Done!

iluwatar:
@inbravo took another look. Here are my comments:
- `DataMapperException` defines additional constructors that are never used. These should be removed.
- You should add `data-mapper` as a module to the main [pom.xml](https://github.com/iluwatar/java-design-patterns/blob/master/pom.xml) Look for `<modules>` declaration and see how the other modules have been added.
- (optional improvement) Since this is a database related pattern it would be great to see actual real database underneath.

Please comment when you're ready for another look.

inbravo:
Parent info is required, i am giving this, please review, build is failed

	<parent>
		<groupId>com.iluwatar</groupId>
		<artifactId>java-design-patterns</artifactId>
		**<version>1.7.0</version>**
	</parent>

iluwatar:
Looks like you have the wrong version number. Try this:
```
<parent>
    <groupId>com.iluwatar</groupId>
    <artifactId>java-design-patterns</artifactId>
    <version>1.12.0-SNAPSHOT</version>
</parent>
```

inbravo:
- Unused constructors from `DataMapperException` discarded
- Main build file updated for addition of new module
- My initial cut was based on DB based assumption; you are absolutely correct it should be DB based, let me see if i can do this ...

inbravo:
This pattern is ready for release; you can plan out ....

iluwatar:
@inbravo I've merged the initial implementation. Looking forward to the real database implementation though.

Issue #418, opened 2016-04-06
Thread-local singleton


iluwatar:
Add to Singleton example: https://dzone.com/articles/singleton-implementation-1

YusufGandhi:
Hi,

I'm Yusuf and am new to this project. I'm currently trying to learn to contribute something. Willing to work on this enhancement, if I may, as the first baby step to contributing to this project in the long run.

Please let me know.

iluwatar:
@YusufGandhi welcome! I'll change the issue labels to indicate you're working on this. Please don't hesitate to ask here or in [Gitter](https://gitter.im/iluwatar/java-design-patterns) if you need help.

YusufGandhi:
@iluwatar Thanks for the prompt reply! I'll definitely ask for help if required.

geniusgeek:
@YusufGhandi @iluwatar, ThreadLocal singleton is not necessarily a
singleton pattern, it is a concurrency resource management pattern. Take a
look at Pattern Oriented Software Architecture by D. Schmidth, This pattern
has been adopted into Java for resource management. Please dont add this as
a Singleton, It was implemented as a singleton so that it can be both
threadsafe, process safe and memory safe.

Cheers

iluwatar:
@geniusgeek I see your point. However, it is inherent to Singleton to take a stand against concurrency. The most basic version creates the static instance eagerly providing thread-safety. The lazily initialized Singletons employ various tricks to be safe when the instance is accessed concurrently. Isn't thread-local Singleton just one more way to solve this?
Also, take a look at related Multiton pattern https://github.com/iluwatar/java-design-patterns/blob/master/multiton/src/main/java/com/iluwatar/multiton/Nazgul.java how it ensures thread-safery with ConcurrentHashMap.

geniusgeek:
All i am saying is that ThreadLocal pattern is not necessarily a Singleton
pattern, it is one of the Concurrent Pattern that ensures shared memory and
shared process pattern. The reason why it is implemented using singleton
pattern is for data consistency. So it is wrong to use it as a Singleton
pattern

npathai:
@iluwatar I second @geniusgeek here. It does not feel right to call thread locals singleton. Though I don't have arguments against that as well. I will figure it out and leave a comment.

iluwatar:
The example is not entirely clear how it relates to Singleton. How about if we write it this way:

```
public final class President {

    // Instance per thread
[...]

PR #422, opened 2016-04-12
Issue #418: Adding additional example of singleton design pattern

YusufGandhi:
Hi,

Please review this commit and let me know if there's something that I left out.

Please be mindful that this is my first baby step to contributing to this project. I might've done something something wrong. So, looking forward to your suggestion.

Cheers.


iluwatar:
The pull request seems to contain only one link whereas I would prefer to add a full code example. However, before we take this any further let's see if we can sort out the indifferences in #418 discussion.

YusufGandhi:
Yes, please resolve the indifferences first and I'll update my commit after that.

iluwatar:
@YusufGandhi can you add the thread-local implementation as discussed in issue #418

Issue #297, opened 2015-11-22
API Gateway pattern


iluwatar:
The example should demonstrate how to use API Gateway in microservices architecture as described here: http://microservices.io/patterns/apigateway.html

TylerMcConville:
I can work on this issue.

iluwatar:
:+1: Great, please go ahead @TylerMcConville

npathai:
@iluwatar @TylerMcConville I would suggest actually using Micro services architecture by deploying various services in separate docker or some other container and letting the API gateway communicate with appropriate service based on device type.

I agree this would be tedious but it would actually represent how pattern is utilized in production. We will need to figure out how to run docker instances while unit testing. Maybe we can mock them out or something. 

Let me know what you think about this.

npathai:
A high quality link: https://www.nginx.com/blog/building-microservices-using-an-api-gateway/

TylerMcConville:
@npathai Sure, I can try to accomplish that. I'll investigate how we would be able to run the Docker instances while unit testing and see what I can come up with.

npathai:
@TylerMcConville We don't necessarily need docker instances in unit testing. We can abstract the Services with interfaces and replace them with our mocks while unit testing.

TylerMcConville:
Just a status update: I've got this pattern code complete, I'm just working on documentation, diagrams, and test coverage.

npathai:
@TylerMcConville That's great :+1:

PR #427, opened 2016-04-24
#297 Create API Gateway pattern

TylerMcConville:
This pattern showcases an API Gateway calling 2 different microservices - one for images, and one for prices.

I created the image and price microservices as submodules of the API Gateway. I needed them to be their own distinct JARs, but also keep them a part of the API Gateway pattern. If there is a better way to do this, please let me know.

The API Gateway and microservices run with Spring Boot. To run them, first build each one by running `mvn package` from the appropriate folder (the api-gateway, price-microservice, or image-microservice folder). That command should create a JAR file in the "target" folder. Run `java -jar target/jar-name.jar` - that should start Spring Boot. You will need to do this for the API Gateway, the Price microservice, and the Image microservice.

Once both microservices and the API Gateway are running, you can test it out by hitting "localhost:50004/desktop" or "localhost:50004/mobile".


iluwatar:
@TylerMcConville in general, it looks really good and works as you instructed. Here are my review comments:
- License headers are missing from some of the files. You can autogenerate them with `mvn clean install`
- In `README.md` I would assign `Architectural` category and `Difficulty-Intermediate` and `Spring` tags
- At the moment the microservices submodules are not built when api-gateway module is built.
- Please move the dependencies to the parent `pom.xml` and only reference them in api-gateway `pom.xml`

iluwatar:
That's it, you have my review remarks. Please comment once you've addressed them and I'll take another look.

TylerMcConville:
@iluwatar I have implemented all of your feedback. The structure has changed a bit: there's a new submodule, api-gateway-service, that contains the code to run the API Gateway itself. The api-gateway parent module now has a packaging type of "pom" and will build all 3 submodules when it is built.

So the updated instructions to run it are:

- Run `mvn package` from the parent module (api-gateway)
- From each of the 3 submodules, in the `target` directory, run `java -jar jar-name.jar`

iluwatar:
@TylerMcConville look perfect! Thanks a lot for the contribution :+1:

iluwatar:
@all-contributors please add @TylerMcConville for code

PR #431, opened 2016-05-29
Renamed HeroBuilder to Builder

VarunVats9:
To make the builder more readable and meaningful, HeroBuilder should have been renamed to Builder, as mentioned in the comments as well. So I did that. I would appreciate it if somebody would review it. Thanks!


iluwatar:
@VarunVats9 thank you, I think this improves the example!

Issue #296, opened 2015-11-22
Microservices architectural pattern


iluwatar:
The example should demonstrate the basic cases how to compose microservices together as described here: http://blog.arungupta.me/microservice-design-patterns/

protopapa:
Hi,

I'm Anastasia . I would like to work on this issue of it is OK. 
Waiting to hear back.

Thanks and Cheers.

iluwatar:
Hi there! You're welcome to work on it. I'll update the issue labels.

protopapa:
Thanks!!!

npathai:
@iluwatar Don't we need splitting of tasks here? Link explains 4-5 different patterns. Are all going to be implemented as one PR?

neonds:
Hi! Maybe I can help in someway.

iluwatar:
@npathai @protopapa as I see it this is a single architectural pattern and there are different flavours to it. In my opinion this issue could implement the basic case and explain the others.

I agree this is a big pattern to implement and would prefer to discuss the best way to implement it before we dig too deep into it.

protopapa:
@iluwatar @npathai As I see it is for sure one pattern with 3 variations. For the last two listings I would like to research and study it a bit more before forming an opinion.

neonds:
I agree with Idea to use the most simplest architecture. I was in a project where we use Repository and Service patterns, with Spring restfull and spring security with Jason Web Tokens (jwt) using DTO pojos . I think that is a good idea use a real example with a complete stack of layers.

protopapa:
Hello! 
I believe that the simplest way to implement this pattern is, as it described on the Aggregator Microservices pattern. I also believe that the Proxy Pattern , Chained Microservices Pattern and Branch Microservices Pattern is a variation of the Aggregator. The Shared Data Microservices Patter and the Asynchronoys Pattern solve a different problem and if implement should be in a different task. 
I am waiting for your feedback. Cheers !

iluwatar:
Hi @protopapa I think you have the right interpretation. For an example, you can look at the excellent [API Gateway pattern](https://github.com/iluwatar/java-design-patterns/tree/master/api-gateway) from @TylerMcConville

protopapa:
Hello again @iluwatar  and @npathai ,
[...]

PR #435, opened 2016-06-23
implementation of microservices architectural pattern #296

protopapa:


iluwatar:
Can you send me instructions how to run the example?

protopapa:
Yes of course! This made me find a bug also! 

So we have three micro-services. From the base directory of each service run : mvn spring-boot:run 
This will start each service in the port that is defined in the resources file. 
Inventory microservice: 51516
Information micro service: 51515
Aggregator service : 50004
Finally in the broswer you hit localhost:50004/product, and you get a JSON with the product , which is assembled using data from the other microservices.

iluwatar:
@protopapa it looks good and works as advertised. Well done! :+1:

Issue #463, opened 2016-07-17
Chain of responsiblity pattern - javax.servlet.Filter as real example is missed


pnowy:
I think that on the chain of responsibility pattern on the section real word examples the most important java real word example is missed.

I suggest to add there **javax.servlet.Filter** as the best patter for the people who meet this patter the first time.

I could create the pull request if it is needed.

iluwatar:
Thanks @pnowy Please create a pull request and we'll review it.

pnowy:
Pull request has been created!

PR #464, opened 2016-07-17
#463 Added javax.servlet.Filter to real world examples intercepting filter

pnowy:
Pull request regarding issue #463 (extended chain real word examples section)

### DIFF
--- a/chain/README.md
+++ b/chain/README.md
@@ -26,7 +26,6 @@ Use Chain of Responsibility when
 
 ## Real world examples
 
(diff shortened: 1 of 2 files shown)


iluwatar:
In Chain of Responsibility pattern processing objects form a chain and process incoming commands. The command is processed by the first processing object and if it doesn't handle the command it passes it on to the next processing object. This continues until the command is processed by one of the processors or the chain is drained (no more processors).

Intercepting Filter pattern is structurally similar. Filters form a chain and then process incoming requests. What distinguishes the pattern from Chain of Responsibility is its behavior. Instead of just one processing object each filter always has a chance to process the request. That is, the request always passes through the whole filter chain.

On this basis, I would say that the links provided in this pull request are examples of Intercepting Filter rather than Chain of Responsibility. I will gladly accept this if you update the pull request this way.

References
https://en.wikipedia.org/wiki/Chain-of-responsibility_pattern
https://en.wikipedia.org/wiki/Intercepting_filter_pattern

pnowy:
You are absolutely right. The pull request has been fixed.

Issue #459, opened 2016-07-08
Borg design pattern


iluwatar:
http://python-3-patterns-idioms-test.readthedocs.io/en/latest/Singleton.html
https://www.quora.com/How-do-I-implement-a-Borg-design-pattern-in-Java

Crossy147:
I will work on this pattern tomorrow.

iluwatar:
Great, please go ahead.

Crossy147:
As far as I see, there already exists a monostate pattern which actually is a borg pattern
[Monostate](https://github.com/iluwatar/java-design-patterns/tree/master/monostate)
What should i do then?

iluwatar:
Well spotted. Borg seems to be an alias for Monostate. It's also said here: http://c2.com/cgi/wiki?MonostatePattern

We should add "Also known as" section with Borg to https://github.com/iluwatar/java-design-patterns/blob/master/monostate/README.md. Would you like to do it?

Crossy147:
yes, i will add this alias today, and open pr to close this issue.

PR #471, opened 2016-07-26
#459 added known alias for monostate pattern

Crossy147:


iluwatar:
Thank you for working on this @Crossy147 :+1:

PR #472, opened 2016-07-26
Applied skeletal implementation to interface migration automated refactoring

khatchad:
This is a semantics-preserving automated refactoring that migrates existing method implementations in classes to the corresponding implemented interface methods as `default` methods. The tool **does not** add new code; it only rearranges *existing* code.

- We are evaluating a research prototype automated refactoring Eclipse plug-in called [Migrate Skeletal Implementation to Interface](https://github.com/khatchad/Migrate-Skeletal-Implementation-to-Interface-Refactoring). We have applied the tool to your project in the hopes of receiving feedback.
- The approach is very conservative. Specifically, it only performs a migration if the target interface is unambiguous. That may mean that not all changes that can be made were made. Please feel free to continue the refactoring manually if you wish.
- We only migrated methods declared in abstract classes. We believe such methods to be likely suitable default methods in corresponding interfaces.
- The source code should be semantically equivalent to the original.

Thank you for your help in this evaluation! Any feedback you can provide would be very helpful. In particular, we are interested if *each* of the proposed changes are helpful or not.

### DIFF
--- a/abstract-document/src/main/java/com/iluwatar/abstractdocument/AbstractDocument.java
+++ b/abstract-document/src/main/java/com/iluwatar/abstractdocument/AbstractDocument.java
@@ -22,12 +22,8 @@
  */
 package com.iluwatar.abstractdocument;
 
-import java.util.List;
 import java.util.Map;
 import java.util.Objects;
-import java.util.Optional;
-import java.util.function.Function;
-import java.util.stream.Stream;
 
 /**
  * Abstract implementation of Document interface
@@ -50,13 +46,6 @@ public abstract class AbstractDocument implements Document {
   @Override
   public Object get(String key) {
     return properties.get(key);
-  }
-
-  @Override
-  public <T> Stream<T> children(String key, Function<Map<String, Object>, T> constructor) {
-    Optional<List<Map<String, Object>>> any = Stream.of(get(key)).filter(el -> el != null)
-        .map(el -> (List<Map<String, Object>>) el).findAny();
-    return any.isPresent() ? any.get().stream().map(constructor) : Stream.empty();
   }
 
   @Override
--- a/abstract-document/src/main/java/com/iluwatar/abstractdocument/Document.java
+++ b/abstract-document/src/main/java/com/iluwatar/abstractdocument/Document.java
@@ -22,7 +22,9 @@
  */
 package com.iluwatar.abstractdocument;
 
+import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.function.Function;
 import java.util.stream.Stream;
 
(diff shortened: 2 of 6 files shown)


iluwatar:
Here's my feedback on the refactoring:

- The changes seem to be technically correct
- The question is why would one want to perform such a refactoring
- Default methods are a language tweak to allow interface evolution
- Default methods are not the new way to write abstract classes
- Default methods are quite limited when compared to abstract classes (cannot be final or synchronized, always public visibility, interfaces cannot contain fields)

So while the pull request could be accepted without breaking anything, I fail to see the benefit. Technically the tool seems to work as specified.

khatchad:
Hi @iluwatar. Thank you for this valuable feedback. You have raised some good questions. I have listed my responses below:

> - The question is why would one want to perform such a refactoring.

We have a [work-in-progress paper](http://websupport1.citytech.cuny.edu/Faculty/rkhatchadourian/media/Towards%20Improving%20Interface%20Modularity%20in%20Legacy%20Java%20Software%20through%20Automated%20Refactoring.pdf) that outlines some of the benefits. There's also a [corresponding presentation](http://www.slideshare.net/khatchad/towards-improving-interface-modularity-in-legacy-java-software-through-automated-refactoring-59573828) (see slide 22 for motivation). In a nutshell, there's certainly more motivation for performing the refactoring in the case that we are able to completely *eliminate* the skeletal implementation classes. That way, extending classes are now free to extend other classes, etc. However, in your particular situation, the tool was not able to do so, perhaps due to the restrictions you mentioned or other kinds of precondition failures. 

In this "partial" case, there is still motivation to perform the refactoring. For example: 

- There are less method declarations; only one is needed in the interface rather than having one in the interface and one in the skeletal implementation class.
- There could be unintentional differences between the method header in the class and the method declaration in the interface. For example, there could be an annotation mismatch between the two, or javadoc inconsistencies. Having only one method declaration could help avoid such problems.
- Not *all* classes may need all of the skeletal implementation class' methods. If they just need the ones that are provided as default methods of the interface, they can forgo extending the skeletal implementation class.
- It may make the interface easier to maintain if evolution does occur as the developer may not need to perform simultaneous edits between the skeletal implementation class and the interface (for example, adding or changing a new method-level annotation, or altering the default implementation).
- There's also a self-documentation benefit as developers considering implementing the interface can see the intended "default" implementation right inside the interface.
- Classes implementing multiple interfaces cannot extend multiple skeletal implementation classes. In this case, they may need to duplicate some of the code found in a skeletal implementation class they cannot extend. But, if those interfaces have *default* methods, it is possible that they can forgo this duplication.

> - Default methods are a language tweak to allow interface evolution.
> - Default methods are not the new way to write abstract classes.

Indeed, the original reason for default methods is to allow for interface evolution. Part of this research is to assess whether they *can* be used in lieu of skeletal implementation class methods, and, if so, how much so.

> - Default methods are quite limited when compared to abstract classes (cannot be final or synchronized, always public visibility, interfaces cannot contain fields).

This is part of the reason why not *all* methods were migrated to interfaces as default methods. It is part of the tools job to determine which methods are allowable. Skeletal implementation methods in abstract must always be publicly visible if they implement interface methods because they cannot reduce visibility (interface methods are always public). Indeed, interfaces cannot contain fields, but they can call mutator and accessor methods (get and set).

Thank you again for your consideration and feedback. You raise some great questions. Please let me know if you have any additional questions!

iluwatar:
@khatchad thank you for the exhaustive answer. You pointed out many benefits but after consideration I think they are not enough to justify the refactoring (in this case). I wish all the best for you and the tool, however!

khatchad:
Thank you for your consideration, @iluwatar.

PR #475, opened 2016-08-02
remove unnecessary statement

haoyaqi:
remove unnecessary statement in singleton/src/main/java/com/iluwatar/singleton/ThreadSafeDoubleCheckLocking.java


npathai:
@haoyaqi The statement is not unnecessary.

This is how it works, Lets go line by line: 

```
ThreadSafeDoubleCheckLocking result = instance;
if (result == null) { //first check
```

First the instance is assigned to result to see if it is already initiated. If the result is null which means it is not initialized. So we need to initialize it.

`synchronized (ThreadSafeDoubleCheckLocking.class) {`

Now because this method will be used in multi-threading context, we synchronize on `.class` which ensures only one thread can be inside the critical zone (i.e. area where we create new instance). 

`result = instance;`

Now once it is ensured that only one thread will be inside this zone we again assign `instance` to `result` to check because some other thread might have initialized it while current thread was blocked to enter the critical zone.

```
if (result == null) { //double check
        instance = result = new ThreadSafeDoubleCheckLocking();
}
```

_Double check_ if the `result` is still `null` which means now that instance is really not initialized so we create a new instance and assign it to `instance`, otherwise some other thread created an instance and assigned it to `instance` while current thread was blocked on entering the critical zone. 

So the line is really not unnecessary but rather half the crux of this pattern. Thanks for your input though :+1:

haoyaqi:
thank you, i learned !

PR #489, opened 2016-08-18
added taste to immutable stew

nahoyman78:
added taste() function to immutable stew


npathai:
Sorry but I didn't get how it adds value to existing example

iluwatar:
The change would make `ImmutableStew` mutable. Also, I second to @npathai that we are not adding value here to the example.

PR #491, opened 2016-08-26
Minor code changes

alexsomai:
I have made minor code changes to the `singleton` module
 - Remove one extra space from `public static ...`
 - Change access level to `private` for `INSTANCE` (this is not strictly related to the pattern, but I think that it makes no sense to have the field public).


npathai:
Thanks for your contribution @alexsomai :+1:

PR #493, opened 2016-09-06
Added example of proxy pattern - Announcer from JMock Demo

npathai:
I recently found out a new application of Proxy pattern. `Announcer` tracks the observers of a particular type and internally keeps a proxy that represents the group of observers. Whenever an announcement is made using the `announce()` method, it returns a proxy instance of the observer type and announcement can be made like a composite object.

It would be great if we can add this sample as well.

If the example is valuable enough then:
- [ ] Complete the javadocs
- [ ]  Update website
- [ ] Pass checkstyle


iluwatar:
@npathai it is a good starting point for a [dynamic proxy](http://docs.oracle.com/javase/8/docs/api/java/lang/reflect/Proxy.html) example. I feel that we should explain what are the use cases for dynamic proxies. In the example we have only one `Endpoint` so it is not entirely clear why the design ended like this.

iluwatar:
Closed due to inactivity

PR #506, opened 2016-10-15
Cosmetic fixes in async-method-invocation

dzmitryh:
There are 2 changes:
- Logic blocks optimized;
- Unused import removed.


iluwatar:
Thank you @dzmitryh for the improvement.

Issue #502, opened 2016-10-06
Logging framework


iluwatar:
Instead of printing output to `System.out` we should use a logging framework such as Log4j to allow redirecting and filtering of output. In some cases this also eases testing. The refactoring should be applied across all the patterns.

dbryla:
I'll handle this. :) I propose to go with SLF4J API and Logback implementation.

iluwatar:
@dbryla the proposal is ok for me, you are good to go.

dbryla:
Should I leave System.out in applications with tests based on StdOut or should I rewrite tests?

iluwatar:
We want to migrate the tests also. If it causes trouble we can deal with it later.

dbryla:
PR: https://github.com/iluwatar/java-design-patterns/pull/515

Build on travis failed:

> The log length has exceeded the limit of 4 Megabytes (this usually means that test suite is raising the same exception over and over).

> The build has been terminated.

Can we increase limit?

PR #515, opened 2016-11-04
#502 Added logging framework to repository

dbryla:


iluwatar:
@dbryla the Travis build is failing with message

```
The log length has exceeded the limit of 4 Megabytes (this usually means that test suite is raising the same exception over and over).

The build has been terminated.
```

I cannot accept the pull request while the build is red.

Do you have recommendation how to deal with this?
* Maybe there is some way to lift this limit in Travis
* Maybe we can limit the output from the build by for example setting the logging level higher

dbryla:
Build is green.

iluwatar:
Thank you so much for the hard work @dbryla This is a great improvement :+1:

PR #519, opened 2016-11-13
Add hexagonal architecture presentation

iluwatar:
As part of issue #197 I made a presentation about Hexagonal Architecture pattern with [remark.js](https://github.com/gnab/remark). If this is ok, I thought we could use similar method for all the pattern presentations.


iluwatar:
@npathai @markusmo3 please review.

markusmo3:
Heya there,
had a quick look through and noticed some things.

* Is the presentation intended to be a readme broken up into smaller pieces? Or more like a blogpost? (I'm already thinking about how we could generate those presentation automatically >.<)
* Maybe it would be nice to also show the class diagram? Im not sure what the usage for such a presentation is, to just show the existence of such pattern or to explain it in detail?
* The slides are kinda neutral, a little "clinical" (thats a german saying, im not sure it conveys the meaning) Does remark.js have themes?

I think of presentations as a more personal thing. Having one is nice to start from, but i wouldn't want to just present a direct copy to a crowd.

All in all, nice work. I didnt even know such a nice thing as remark.js existed before :+1:

iluwatar:
Thank you for the comments.

>>> Is the presentation intended to be a readme broken up into smaller pieces? Or more like a blogpost? (I'm already thinking about how we could generate those presentation automatically >.<)

I thought about it while doing this, but I think the presentations need a bit different angle and can't be autogenerated from the content we already have. So yes, it has some bits from the `README.md` and then some new content.

>>> Maybe it would be nice to also show the class diagram? Im not sure what the usage for such a presentation is, to just show the existence of such pattern or to explain it in detail?

Good point. I agree that for some patterns showing the class diagram makes sense but for this large architectural pattern I chose to leave it out.

>>> The slides are kinda neutral, a little "clinical" (thats a german saying, im not sure it conveys the meaning) Does remark.js have themes?

Yes, I used the default styles and agree that it looks clinical. I'll look into this.

Tschis:
@iluwatar did you do it all by hand, or have you automated any part of it?
I thought I could help, but if it is all by hand then, as @markusmo3 suggested, we might get outdated presentations while the project continues to expand and update.

iluwatar:
At the moment I'm not working on this but we are accepting pull request for presentations. We have selected remarkjs as the technology but otherwise it is manual work. I believe it can't be automated for good results. See https://github.com/iluwatar/java-design-patterns/wiki/01.-How-to-contribute for instuctions.

PR #520, opened 2016-11-14
Update App.java

thomasoss:
The relationship between decorator and target (in the code example between Troll und SmartTroll) ist not a composition as described in the code comment lines of class App. It's a aggregation. In a composition both instances canncot live without the other one. But the target (Troll) could live without the decorator (Smart Troll). Changed "composition" to "aggregation" in the comment lines. Code lines are unchanged.
Greetings Thomas


iluwatar:
Thank you @thomasoss for this fix :+1:

Issue #77, opened 2015-06-23
Thread local storage pattern


iluwatar:
https://en.wikipedia.org/wiki/Thread-local_storage

thomasoss:
I would like to deal with this
Greetings Thomas

iluwatar:
Yes, great @thomasoss, the issue is yours!

thomasoss:
The coding is done and I also worked through the Checkstyle Pluging messages. But for using  "ObjectAid UML Explorer for Eclipse" the purchase of a license is obligatory. Is there another way to create the UMLs in the format you need?

mrcrow85:
@thomasoss The license is only needed if you want to use the extra diagrams, the free version already has the class diagrams feature which is the diagram you need.

thomasoss:
@mrcrow85 thank you. Worked.
Now the UML is created. Looking into the existing etc folders is see xyz.urm.puml-files too. Can ObjectAid create this file. I couldn't find out how.
In the readme.md there is a pumlid: Where does this come from?
Greetings Thomas

iluwatar:
We are in the process of switching the tool used to create the UML diagrams. This is discussed in https://github.com/iluwatar/java-design-patterns/issues/190. If you want, you can try the steps proposed by @markusmo3 to generate the pumlid or optionally just leave the pumlid empty for now.

thomasoss:
... ok - the tls.urm.puml was generated by the Maven build. Because my environment is configured for Gradle I had to do the Maven build in a second Eclipse project to which I copied the source code files. And I oversaw the puml created in the etc folder.

thomasoss:
@iluwatar and @markusmo3 
creation of the pumlid worked fine with your instructions. Thank you

PR #529, opened 2016-12-16
Thread Local Storage issue #77

thomasoss:
Example for using a ThreadLocal variable in a Runnable object. To show the problems when not using the Thread Local Storage pattern a second app named AppUgly is added. This second app does not use the Thread Local Storage pattern and produces exeception and / or bad results. For AppUgly no JUnit tests are provided because these would neither surely fail nor surley succeed. Both apps are described in the code (javadoc).
I could not find a way to check the rendering of the READMD.md.


iluwatar:
Here are my initial review comments:
- tls module is missing from the parent `pom.xml`.
- The example is not very meaningful at the moment. Could we think of an example with a story to make it more interesting? See the other patterns for examples.
- Use of plain `Thread`s is kind of an antipattern. Prefer using `ExecutorService` instead.
- I would leave out the ugly example. The consequences could just be explained. Now it is kind of copy-paste of the main example.
- I would separate the `AppTest` and other tests to make it uniform with the other patterns.

@thomasoss please comment on this thread once the comments have been addressed.

thomasoss:
Thank you for your comments. I'm a little bit busy at the moment but I will deal with them within the next weeks. Have good start for 2017.

thomasoss:
Hello

-  tls is added to the parent pom.xml. I copied the code of the current version of pom.xml the 29th January 2017. Despite this beside the addition of tls two other changes are shown in the pull request. The line with the version and some eof marker. Maybe I'm a little bit unexperienced using github.
- The example is unchanged. I don't have time at the moment to change it to another example. I can do it not before April because of a 3 week holiday trip and a 2 week business trip until the end of March.
- ExecutorService is used now
- ugly example is left out. For testing the problems arising with not using Thread Local Storage interested persons can comment out and comment in some marked pieces of DateFormatCallable
- AppTest is separated

The pumlid I created last time was not correct. This time no pumlid is enclosed due to your proposal in december regarding the pumlid

tls/pom.xml
```diff
@@ -0,0 +1,43 @@
...
+    THE SOFTWARE.
+
+-->
+<project xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns="http://maven.apache.org/POM/4.0.0"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>com.iluwatar</groupId>
+    <artifactId>java-design-patterns</artifactId>
+    <version>1.14.0-SNAPSHOT</version>
```
iluwatar: Should be 1.15.0-SNAPSHOT

thomasoss:
The code commented out is by purpose to allow running the example without thread local storage and see the results. Please view it. But I can remove it of course.
I changed  to <version>1.15.0-SNAPSHOT</version> in the tls/pom.xml. I could verify the build failure with the current version of the parent pom.xml und the "1.14-SNAPSHOT" entry in the tls/pom.xml in my local environment. After changing it to 1.15 the build was successful in my local environment. I hope it will do in the target environment too.

iluwatar:
@thomasoss the build is still failing. You can see the build log by clicking the `Details` next to Travis icon.

thomasoss:
@iluwatar I could reproduce the fail. I used the same mvn call as shown in the build log (mvn install -DskipTests=true -Dmaven.javadoc.skip=true -B -V -e). I corrected the error found by the maven-pmd-plugin and in my environment the maven build is now successful. I hope it will do so in the target environment.

iluwatar:
@thomasoss this looks good. Thank you for the contribution :+1:

iluwatar:
@all-contributors please add @thomasoss for code

PR #530, opened 2016-12-18
Sonar quality bugfix and minor refactor

Valdar-Soft:
Sonar bug fixes using Yoda condition in equals expression when comparing String literal with String object. 
Using try-with-resources if we use Scanner to close the underlying stream is a good practice to handle resources.
Minimal refactor.


galistvan:
awesome :)

iluwatar:
@Valdar-Soft thank you for this improvement! :+1:

PR #532, opened 2016-12-24
Adding initialization-on-demand idiom and noninstantiable class instead of interface constant idiom

leogtzr:
Adding noninstantiable class instead of interface constant idiom.


iluwatar:
In the Singleton example [IvoryTower](https://github.com/iluwatar/java-design-patterns/blob/master/singleton/src/main/java/com/iluwatar/singleton/IvoryTower.java) demonstrates eagerly initialized instance. In this pull request you have changed it implement [Initialization-on-demand holder idiom](https://en.wikipedia.org/wiki/Initialization-on-demand_holder_idiom). However, this has already been demonstrated in [InitializingOnDemandHolderIdiom](https://github.com/iluwatar/java-design-patterns/blob/master/singleton/src/main/java/com/iluwatar/singleton/InitializingOnDemandHolderIdiom.java).

iluwatar:
The second change concerning [CustomerSchemaSql](https://github.com/iluwatar/java-design-patterns/blob/master/dao/src/main/java/com/iluwatar/dao/CustomerSchemaSql.java) doesn't add any value since variables in interfaces are static and final by default.

leogtzr:
I know they are static and final by default, the problem with the interface is that somebody can implement from it. We don't want that. Those classes were not designed to be instantiated. I am enforcing noninstantiability by using this idiom:

- interface -> class
- include a private constructor.

iluwatar:
Ok, after reading more about the [constant interface](https://en.wikipedia.org/wiki/Constant_interface) idiom I tend to agree that it is better style to declare final class instead. If you update the pull request and comment on this thread I can accept it.

leogtzr:
Added a little description, thanks :)

leogtzr:
Added a few more changes, found that several java classes were using references to implementations instead of using abstractions/interfaces. Please review :D @iluwatar

iluwatar:
@leogtzr looks good. Still the changes concerning `IvoryTower` need to be reverted for the reasons I mentioned before.

leogtzr:
@iluwatar done, I have reverted the changes for `IvoryTower.java`, please take a look :)

iluwatar:
@leogtzr thank you for the improvements!

PR #546, opened 2017-03-11
Converter

Crossy147:
My purpose of a Converter being a generic class instead of an interface was to enable very simple creation of an instance, providing only two converters as Java8 lambdas - very clean solution (example usage here: https://github.com/iluwatar/java-design-patterns/compare/master...Crossy147:converter?expand=1#diff-a47ad9e33423d70a452559b5562b8548R44)


Crossy147:
This implementation covers the issue #546

Crossy147:
I've resolved pom conflict, taking into account the merged PR 
@iluwatar - It's ready for your review :)

iluwatar:
@all-contributors please add @Crossy147 for code

PR #565, opened 2017-04-12
Update ObjectPool.java

kapinuss:
available.size() < 0 is impossible, so it has sense only in case available.size() = 0, and it means that collection is empty, besides function isEmpty() is faster, than checking of the size.


iluwatar:
Thanks @kapinuss :+1:

PR #568, opened 2017-04-26
Fix incorrect URL

brandon-d-mckay:


iluwatar:
Thanks for the fix @brandon-d-mckay :+1:

PR #571, opened 2017-04-30
Fix Block 3 Case in Feature Toggle Design Pattern

ytian90:
The original code use PropertiesFeatureToggleVersion(), but for Block 3 case, it should use TieredFeatureToggleVersion().


iluwatar:
@ytian90 thanks for spotting and fixing this!

PR #583, opened 2017-05-25
Updated ThreadSafeLazyLoadedivoryTower

shaiktaj:
Added null check in private constructor to prevent instantiating by Reflection call


iluwatar:
@shaiktaj  this pull request is about to be closed due to inactivity. Please comment if you are still working on this.

shaiktaj:
@iluwatar The IllegalStateException is being thrown right?

iluwatar:
Thanks @shaiktaj it looks good :+1:

igorcadelima:
It's still possible to create many instances of the class using only reflection.

iluwatar:
It is not a complete solution, but a step in the right direction. See https://stackoverflow.com/questions/6994393/singleton-how-to-stop-create-instance-via-reflection/6994437

PR #593, opened 2017-06-14
Refactored Function<Integer,T> to IntFunction<T>

Harshrajsinh:
This is another similar change that found. This will greatly improve the performance as the `for `loop runs `128*1024 times` and boxing happens in every loop.


iluwatar:
Thank you @Harshrajsinh for this improvement! :+1:

Issue #587, opened 2017-06-02
SonarQube reports bugs


iluwatar:
The project is automatically scanned with SonarQube. The static code analysis reports several blocker and critical level issues that should be fixed.

https://sonarcloud.io/project/issues?id=com.iluwatar%3Ajava-design-patterns&resolved=false&severities=BLOCKER%2CCRITICAL&types=BUG

mookkiah:
Can I work on this issue?

iluwatar:
@mookkiah ok

mookkiah:
@iluwatar In com.iluwatar.dao.DbCustomerDao.getAll() we are returning stream so the resources (connection, prepared statement and resultset) are should be available until stream is used by the consumer. If we close in finally, we are getting error "The object is already closed".
I think we should just add //NOSONAR to ignore. This is not a blocker as we are closing connection when stream is closed.  Please let me know if you think of another solution.

iluwatar:
@mookkiah the proposed solution is fine

iluwatar:
@mookkiah  are you still implementing this?

mookkiah:
@iluwatar This issue can be closed as there is no other sonarqube issue as per the criteria.

iluwatar:
True, well done @mookkiah :+1:

PR #599, opened 2017-07-22
#587 sonarqube bugs

mookkiah:


intercepting-filter/src/main/java/com/iluwatar/intercepting/filter/Client.java
```diff
@@ -37,20 +35,20 @@
...
- * This is where {@link Filter}s come to play as the client pre-processes the request before being
- * displayed in the {@link Target}.
+ * This is where {@link Filter}s come to play as the client pre-processes the request before being displayed in the
+ * {@link Target}.
  * 
  * @author joshzambales
  *
  */
-public class Client extends JFrame {
+public class Client extends JFrame { // NOSONAR
```
iluwatar: Why is `// NOSONAR` needed here?

intercepting-filter/src/main/java/com/iluwatar/intercepting/filter/Client.java
```diff
@@ -37,20 +35,20 @@
...
 
   private static final long serialVersionUID = 1L;
 
-  private FilterManager filterManager;
+  private transient FilterManager filterManager;
```
iluwatar: Making it transient doesn't introduce any problems?

intercepting-filter/src/main/java/com/iluwatar/intercepting/filter/Target.java
```diff
@@ -42,7 +42,7 @@
  * @author mjoshzambales
  *
  */
-public class Target extends JFrame {
+public class Target extends JFrame { //NOSONAR
```
iluwatar: Why is `// NOSONAR` needed here?

intercepting-filter/src/main/java/com/iluwatar/intercepting/filter/Client.java
mookkiah: Sonarqube reported "This class has 6 parents which is greater than 5 authorized"
Here is the inheritance hierarchy. 
FileSelectorJFrame -> JFrame -> Frame -> Window -> Container -> Component implements ImageObserver, MenuContainer, Serializable

I believe, it is most common to subclass JFrame for creating window. 
Alternate design by following "Use composition over inheritance". I thought it would make more changes and may spoil the brevity of explaining the design pattern. 
Please let me know if you want me to think about the alternate design

intercepting-filter/src/main/java/com/iluwatar/intercepting/filter/Client.java
mookkiah: It did not introduce any problem. 
Alternate solution would be making FilterManager and its parents as Serializable. In this example, we are not transferring Client instance or planning to serialize, So I decided to make it transient to keep it simple. 

Let me know if you prefer marking all classes as Serializable as long as it does not have any non-serialized attributes.

iluwatar:
@mookkiah thanks for the explanations. It looks good to be merged. Well done :+1:

PR #605, opened 2017-08-05
#587 SonarQube reports bugs

mookkiah:
Marking FileLoader (Model) as Serializable, because model is an interface to provide data and it can provide caching behaviour. Presenter should be able to restore the model in any event.
Marking FileSelectorView as Serializable as I think view is made up of some static structure and Serializable data (from model).
Presenter is an middle-man who manages View and Model. Both View and Model will be given to Presenter by the application. It will not make any harm to make presenter as Serializable. It will be useful serialize model and view as it is tied to one another.

Or, we could set //NOSONAR and let the consumer application framework to decide whether to mark as Serializable or not based on their design.

References: 
https://en.wikipedia.org/wiki/Model%E2%80%93view%E2%80%93presenter
https://medium.com/@cervonefrancesco/model-view-presenter-android-guidelines-94970b430ddf

### DIFF
--- a/model-view-presenter/src/main/java/com/iluwatar/model/view/presenter/FileLoader.java
+++ b/model-view-presenter/src/main/java/com/iluwatar/model/view/presenter/FileLoader.java
@@ -25,6 +25,10 @@ package com.iluwatar.model.view.presenter;
 import java.io.BufferedReader;
 import java.io.File;
 import java.io.FileReader;
+import java.io.Serializable;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * Every instance of this class represents the Model component in the Model-View-Presenter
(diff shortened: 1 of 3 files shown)


iluwatar:
@mookkiah I'm not sure which SonarQube issue this PR fixes?

mookkiah:
@iluwatar, This is for issue #587

iluwatar:
@mookkiah yes, I found it now. This https://sonarcloud.io/project/issues?id=com.iluwatar%3Ajava-design-patterns&open=AVfejjKvPl_GJI-SB0k5&resolved=false&severities=BLOCKER%2CCRITICAL&types=BUG%2CCODE_SMELL

iluwatar:
Thanks @mookkiah the fix has been merged :+1:

Issue #348, opened 2016-01-16
Data Transfer Object pattern


iluwatar:
http://martinfowler.com/eaaCatalog/dataTransferObject.html

raviteja1992y:
I started working on this

iluwatar:
Ok, I will update the labels. Don't hesitate to ask if you need support.

iluwatar:
@raviteja1992y are you still on working on this?

raviteja1992y:
@iluwatar  Yeah. I am working on this.

iluwatar:
@raviteja1992y how's it going? Please let us know if you are still implementing this pattern.

iluwatar:
There is no answer, so this issue is free for taking again.

squla:
@iluwatar 
I can do it. But it should be simple example i.e. we get two things from db and we prepare DTO, or maybe i should present usage with rest api ?

gopinath-langote:
@iluwatar 
Is anyone still working on it? I can take it up for development.

iluwatar:
@squla are you still working on it?

iluwatar:
@gopinath-langote no reply from @squla so I guess we can transfer the issue for you to work on.

gopinath-langote:
@iluwatar I am working on this.

iluwatar:
Ok @gopinath-langote

iluwatar:
Puml id missing

gopinath-langote:
@iluwatar Added Piml id to README.md

PR #608, opened 2017-08-11
#348 Data Transfer Object design pattern

gopinath-langote:
Basic version of Data transfer object pattern implementation.


data-bus/pom.xml
```diff
@@ -27,6 +27,10 @@
          xmlns="http://maven.apache.org/POM/4.0.0"
          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
     <modelVersion>4.0.0</modelVersion>
+    <packaging>pom</packaging>
+    <modules>
+        <module>../data-transfer-object</module>
+    </modules>
```
iluwatar: Why is Data Bus pattern affected? Should this change be reverted?

data-transfer-object/src/main/java/com/iluwatar/datatransfer/CustomerClientApp.java
```diff
@@ -0,0 +1,78 @@
...
+   * @param args program argument.
+   */
+  public static void main(String[] args) {
+    List<CustomerDto> customers = new ArrayList<>();
+    CustomerDto customerOne = new CustomerDto("1", "Kelly", "Brown");
+    CustomerDto customerTwo = new CustomerDto("2", "Alfonso", "Bass");
+    customers.add(customerOne);
+    customers.add(customerTwo);
+
+    CustomerResource customerResource = new CustomerResource(customers);
```
iluwatar: This is customer server data initialization, right? Could we move to static block in `CustomerResource` class?

data-transfer-object/src/main/java/com/iluwatar/datatransfer/CustomerClientApp.java
```diff
@@ -0,0 +1,78 @@
...
+
+    System.out.println("Adding customer three}");
+    CustomerDto customerThree = new CustomerDto("3", "Lynda", "Blair");
+    customerResource.save(customerThree);
+    allCustomers = customerResource.getAllCustomers();
+    printCustomerDetails(allCustomers);
+  }
+
+  private static void printCustomerDetails(List<CustomerDto> allCustomers) {
+    allCustomers.forEach(customer -> System.out.println(customer.getFirstName()));
```
iluwatar: Don't use `System.println`, instead add and use a logger as in other examples.

data-transfer-object/src/main/java/com/iluwatar/datatransfer/CustomerDto.java
```diff
@@ -0,0 +1,60 @@
...
+/**
+ * {@link CustomerDto} is a data transfer object POJO. Instead of sending individual information to client
+ * We can send related information together in POJO.
+ * <p>
+ * Dto will not have any business logic in it.
+ */
+public class CustomerDto {
+  private String id;
+  private String firstName;
+  private String lastName;
```
iluwatar: This looks like immutable class. The private members should be declared `final`.

data-bus/pom.xml
gopinath-langote: @iluwatar 👍 . 
Should be reverted. I will do that.

data-transfer-object/src/main/java/com/iluwatar/datatransfer/CustomerClientApp.java
gopinath-langote: @iluwatar To keep is simple yes. We can do that. Rather that looks more appropriate. 
Thanks for suggestions. Let me do that.

data-transfer-object/src/main/java/com/iluwatar/datatransfer/CustomerClientApp.java
gopinath-langote: @iluwatar 👍 Ok. Make sense. I'll do that.

data-transfer-object/src/main/java/com/iluwatar/datatransfer/CustomerDto.java
gopinath-langote: @iluwatar : Sure. Thanks for pointing this out. I'll to the same.

gopinath-langote:
@iluwatar : Added missing class diagram. 

Thanks.

iluwatar:
@gopinath-langote looks good now. Thanks for the contribution! :+1:

PR #609, opened 2017-08-12
Protect against instantiation via reflection

igorcadelima:
Add missing requirements from #583 .

Signed-off-by: Igor C. A. de Lima <igorcadelima@gmail.com>


iluwatar:
@igorcadelima the change looks good. Would it be possible to prove it correct by adding a test case?

iluwatar:
Thanks @igorcadelima :+1:

PR #611, opened 2017-08-13
Issue 587 SonarQube reports bugs - Module DAO

mookkiah:
As getAll method returns a Stream, we cannot close the involved resources (connection, statement and resultset) until the stream is closed by the consumer. So try-with-resources is not an option as per sonarqube’s recommendation. But it is still recommended to close statement and result set. When connection pool used, connection is not closed when close() called. It is just returned to the pool.

Using //NOSONAR to avoid false blocker issue.


iluwatar:
Good job @mookkiah :+1:

PR #616, opened 2017-08-18
#587 SonarQube reports bugs - async-method-invocation

mookkiah:
As recommended in
https://sonarcloud.io/organizations/default/rules#rule_key=squid%3AS2274 
Used while insteadof if - for waiting upon a condition.


iluwatar:
The fix looks good. Thanks again @mookkiah :+1:

PR #617, opened 2017-08-19
#587 SonarQube reports bugs reader-writer-lock and refactor

mookkiah:
Keeping wait() calls with in synchronized block closely to adhere
SonarQube rules.
 
Avoid nested synchronized block by extracting method.

Added writing and reading time to simulate testing to ensure 
1) writers are waiting for globalMutex to be empty 
2) readers to confirm there is no writers.


mookkiah:
@iluwatar, I could resolve this issue by adding NOSONAR comment to make it simple change. But I noticed an opportunity to refactor, follow SonarQube rules and make the App.java to explain the reader writer mutex.

iluwatar:
@mookkiah very nice! Thanks for the additional refactoring :+1:

PR #618, opened 2017-08-22
#348 - Data Tranfer Object : Add Puml id to README.md.

gopinath-langote:
Issue: #348 

Pattern: Data Transfer Object

---
Add a missing Puml id to README.md


iluwatar:
Thanks for the fix @gopinath-langote :+1:

PR #620, opened 2017-08-29
Improved IvoryTower class of Singleton

dharan1011:
Rather than creating final intsance of IvoryTower class, its better to check the nullity of INSTANCE variable then create object else return already instanciated object when `getInstance` method is called.


iluwatar:
The purpose of `IvoryTower` class is to explain eagerly initialized Singleton. The example has other type of Singletons too including the lazy loaded one you are suggesting. Please read the comments in https://github.com/iluwatar/java-design-patterns/blob/master/singleton/src/main/java/com/iluwatar/singleton/App.java

dharan1011:
Opps.. I must have gone throught the repo. Apologies !! :disappointed:

Issue #607, opened 2017-08-09
Searching for patterns doesn't work as expected


mibac138:
When using the search box on the [patterns](http://java-design-patterns.com/patterns/) site I wanted to see the Prototype pattern. 

I started writing:
- `p` - not much changed
- `r` - the same
- `o` - every pattern disappeared
- this continues until `p`
- `p` - 3 patterns appear: `Naked Objects` (how?), `Property` and `Prototype`
- `e` - still 3 patterns appearing

Why does searching act so weird? I suppose this is a bug

iluwatar:
Ping @markusmo3

markusmo3:
Its black magic, i dont understand myself ^^
Anyone is free to try and fix it!

Here some important parts:
* it uses elasticlunr.com
* https://github.com/iluwatar/java-design-patterns/blob/gh-pages/search_index.json
* https://github.com/iluwatar/java-design-patterns/blob/gh-pages/static/js/search.js
* https://github.com/iluwatar/java-design-patterns/blob/98827c9537ef868fda01868b2f6dc91d28b2a36c/pages/patterns.html#L18-L24

iluwatar:
Thanks @markusmo3 looks like nice puzzle :smile:

danielsiwiec:
I took a stab at this and realized that replacing it with Lunr which supports wildcard searches makes it possible to search by shorter substrings.

Additionally, I removed the index boosts on certain fields, as they were never used anyway - they don't affect what is being returned by the query, only the score field in the result items and so far, this field isn't used by this application

iluwatar:
@danielsiwiec see https://github.com/iluwatar/java-design-patterns/issues/653 Any idea what is going on?

PR #625, opened 2017-09-01
Replace elasticlunr with lunr as it allows wildcard searches #607

danielsiwiec:
Addresses issue #607


iluwatar:
Looks like the original issue (https://github.com/iluwatar/java-design-patterns/issues/607) is resolved with this pull request and searching for `prototype` gives only one correct result.

However it seems that some other searches are broken. Testing results for searches
- `caching` gives no results
- `cqrs` gives no results
- `api gateway` gives 9 results

@danielsiwiec can you reproduce this? The search behaves very oddly for me.

danielsiwiec:
@iluwatar I think I finally figured it out - the lunrj stemmer was to blame for this (a component that maps all words added to the index to their base form - e.g. 'caching' -> 'cach'). This was also likely messing up cqrs and these searches should look good now.

For the 'API Gateway' - the reason it shows multiple results is that the other results have 'API' in their content.
We could be more restrictive and either:
a) search only in tags and titles
b) set a threshold for score (each search result comes back with a score, that quantifies how precise the search 'hit' is)

iluwatar:
@danielsiwiec it looks much better now and finds the correct pattern :+1: Still I'm bit confused why it displays multiple results in some cases
- searching for `api` gives 9 results e.g. Aggregator Microservices is one of them, why? It doesn't have `api` in the pattern name.
- searching for `command` gives 2 results, the correct one and CQRS, why?
- searching for `state` gives 14 results (including the correct one), why?

danielsiwiec:
@iluwatar the reason for those additional results to show is that their content (the actual pattern page) contains those keywords. To prevent this, we can disable the content (pattern pages) from being indexed and used in the search.
I think a reasonable approach is to search the names, tags and categories only. Thought?

iluwatar:
@danielsiwiec yes, I think we should disable the content then. The search is more like a filter, not a Google search.

danielsiwiec:
@iluwatar done. Left the title, category and tags to be searchable

iluwatar:
Works perfectly. Thanks @danielsiwiec this is hugely appreciated! :+1:

Issue #456, opened 2016-07-02
Throttling pattern


iluwatar:
https://msdn.microsoft.com/en-us/library/dn589798.aspx

rastdeepanshu:
Can I work on this pattern if no one else is working? Also is there a deadline to complete it?

iluwatar:
Ok @rastdeepanshu go ahead. No strict deadlines. Please ask for help if you are stuck.

rastdeepanshu:
@iluwatar I have created a service which can throttle based on time i.e. a service is accessible for a number of times per second. Any other calls more than the allowed calls are rejected.

Is that something of interest?

iluwatar:
@rastdeepanshu that sounds like a great example.

rastdeepanshu:
@iluwatar I am getting this error while building the project:

`Failed to execute goal com.github.markusmo3.urm:urm-maven-plugin:1.4.4:map (default) on project throttling: /Users/drastogi/java-design-patterns/throttling/etc/throttling.urm.puml
`

I am not sure what needs to be done. Can you tell?

rastdeepanshu:
@iluwatar Awaiting review

iluwatar:
https://docs.microsoft.com/en-us/azure/architecture/patterns/throttling

PR #629, opened 2017-09-07
Throttling pattern

rastdeepanshu:
[#456](https://github.com/iluwatar/java-design-patterns/issues/456)

Created throttling pattern which throttles request of a particular tenant.
The service only allows a specified number of requests for a tenant and rejects the remaining ones.


throttling/README.md
```diff
@@ -0,0 +1,20 @@
...
+folder: throttling
+permalink: /patterns/throttling/
+tags:
+ - Java
+ - Difficulty-Beginner
+ - Throttling
+---
+
+## Intent
+Ensure that a given tenant is not able to access resources more than the assigned limit.
```
iluwatar: Should we talk about client here instead of tenant? And service resources instead of just resources. Check for occurrences elsewhere too.

throttling/src/main/java/com/iluwatar/tls/App.java
```diff
@@ -0,0 +1,85 @@
...
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package com.iluwatar.tls;
```
iluwatar: com.iluwatar.throttling?

throttling/src/main/java/com/iluwatar/tls/App.java
```diff
@@ -0,0 +1,85 @@
...
+   * Application entry point
+   * @param args main arguments
+   */
+  public static void main(String[] args) {
+
+    Tenant adidas = new Tenant("Adidas", 80);
+    Tenant nike = new Tenant("Nike", 70);
+
+    B2BService adidasService = new B2BService(adidas);
+    B2BService nikeService = new B2BService(nike);
```
iluwatar: The relationship between service and tenant here is one-to-one which seems unnatural. The service should have multiple tenants, I think.

throttling/src/main/java/com/iluwatar/tls/App.java
```diff
@@ -0,0 +1,85 @@
...
+
+    Runnable adidasTask = () -> makeServiceCalls(adidasService);
+    Runnable nikeTask = () -> makeServiceCalls(nikeService);
+
+    new Thread(adidasTask).start();
+    new Thread(nikeTask).start();
```
iluwatar: Threads should not be instantiated directly. Instead use `ExecutorService`.

throttling/src/main/java/com/iluwatar/tls/App.java
```diff
@@ -0,0 +1,85 @@
...
+   */
+  private static void makeServiceCalls(B2BService service) {
+    for (int i = 0; i < 500; i++) {
+      try {
+        service.dummyCustomerApi();
+//      This block is introduced to keep the output in check and easy to view and analyze the results.
+        try {
+          Thread.sleep(10);
+        } catch (InterruptedException e) {
+          e.printStackTrace();
```
iluwatar: Should use logger here

throttling/src/main/java/com/iluwatar/tls/App.java
```diff
@@ -0,0 +1,85 @@
...
+        }
+//      It can be removed if required.
```
iluwatar: Remove this comment

throttling/src/main/java/com/iluwatar/tls/B2BService.java
```diff
@@ -0,0 +1,86 @@
...
+import org.slf4j.LoggerFactory;
+
+import java.nio.file.AccessDeniedException;
+import java.util.Timer;
+import java.util.TimerTask;
+import java.util.concurrent.ThreadLocalRandom;
+
+/**
+ * A service which accepts a tenant and throttles the resource based on the time given to the tenant.
+ */
```
iluwatar: The class seems to have multiple responsibilities. It offers the service and handles the throttling. Can we separate the throttling to another class? Maybe time based throttling would be one strategy but it could have others.

throttling/src/main/java/com/iluwatar/tls/B2BService.java
```diff
@@ -0,0 +1,86 @@
...
+class B2BService {
+
+  private Tenant tenant;
+  private int callsCounter;
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(B2BService.class);
+
+  /**
+   * A timer is initiated as soon as the Service is initiated. The timer runs every minute and resets the
+   * counter.
```
iluwatar: Seems to run every second

throttling/src/main/java/com/iluwatar/tls/Tenant.java
```diff
@@ -0,0 +1,64 @@
...
+    this.allowedCallsPerSecond = allowedCallsPerSecond;
+  }
+
+  public String getName() {
+    return name;
+  }
+
+  public void setName(String name) {
+    this.name = name;
+  }
```
iluwatar: The method is unused and should be removed

throttling/src/main/java/com/iluwatar/tls/Tenant.java
```diff
@@ -0,0 +1,64 @@
...
+
+  public int getAllowedCallsPerSecond() {
+    return allowedCallsPerSecond;
+  }
+
+  public void setAllowedCallsPerSecond(int allowedCallsPerSecond) {
+    this.allowedCallsPerSecond = allowedCallsPerSecond;
+  }
```
iluwatar: The method is unused and should be removed

throttling/src/test/java/com/iluwatar/tls/B2BServiceTest.java
```diff
@@ -0,0 +1,51 @@
...
+    for (int i = 0; i < 20; i++) {
+      service.dummyCustomerApi();
+      try {
+        Thread.sleep(100);
+      } catch (InterruptedException e) {
+        e.printStackTrace();
+      }
+    }
+    int counter = service.getCurrentCallsCount();
+    Assert.assertTrue("", counter < 11);
```
iluwatar: Where this magic number 11 comes from?

iluwatar:
Add `AppTest` that runs the example. See the other patterns for an example.

throttling/pom.xml
```diff
@@ -0,0 +1,44 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
iluwatar: The license is incorrectly positioned

iluwatar:
@rastdeepanshu you have my review comments. Please notify when you've implemented the changes.

throttling/src/main/java/com/iluwatar/tls/App.java
rastdeepanshu: B2BService is one service which serves multiple tenants. As I am not dealing with the APIs, one instance of the service deals with one tenant.

throttling/src/test/java/com/iluwatar/tls/B2BServiceTest.java
rastdeepanshu: sleep for 100 ms means counter can go till 10. Hence the magic number :D 
I can do <= 10 if 11 is looking odd.

throttling/src/main/java/com/iluwatar/tls/App.java
rastdeepanshu: I wanted these two tasks with no relation b/w them. ExecutorService will execute them using a combined optimised login. I actually had tried with ES but the output is not what I expected. When the NikeService completes, it just continues with NikeService instead of giving execution time to AdidasService. And vice-versa.

throttling/pom.xml
rastdeepanshu: Can you elaborate. I guess the first line has to be this for pom to work correctly.

throttling/src/main/java/com/iluwatar/tls/B2BService.java
rastdeepanshu: I see

throttling/src/main/java/com/iluwatar/tls/App.java
rastdeepanshu: I get your point now.

rastdeepanshu:
@iluwatar Can you take a look now?

iluwatar:
@rastdeepanshu well done. Thanks for your effort to add this new pattern :+1:

PR #632, opened 2017-09-14
Delete duplicate declared objects:ClubbedTroll

OCNYang:
Trolls should not be re-declared: if I had a troll called Zac, one day I
gave him a weapon.I can not because he now has a weapon called him
Trundle. He is still Zac.


iluwatar:
Good point @OCNYang Thanks for the improvement! :+1:

PR #634, opened 2017-09-15
#587SonarQube reports bugs in promise module

mookkiah:


npathai:
@mookkiah Aha interesting. I will have a look at it.

iluwatar:
@mookkiah thanks for the fix! :+1:

PR #639, opened 2017-09-19
Fix naked objects thesis url

reegnz:
The commit fixes the URL, changing from a non-working URL to an actually
working one.
The URL was copied from the wikipedia page for Naked Objects.


iluwatar:
The fix looks good. Thanks @reegnz :+1:

PR #640, opened 2017-09-21
Fixes Checkstyle problems

dosdebug:
There are few checkstyles problems found in the  which are required to be fixed


iluwatar:
Looks good. Thanks @dosdebug :+1:

Issue #547, opened 2017-03-18
Incorrect history stack handling in command pattern


mihaildemidoff:
Hello!

Thank you for your work.
I found one small problem in command pattern. I agree that code should be light as possible and some details may be ommited, but examples should be correct.
Problem: InvisibilitySpell and ShrinkSpell have incorrect redo and undo mechanism. For example: if you execute ShrinkSpeel 2 times Goblin will never return to Big size.
I think we should add stack of states in commands to properly handle redo and undo operations.

mattj256:
@iluwatar I'd like to work on this.

The Size enumerated type has small, medium, large, and undefined.  It seems that only small and medium are ever used.  Should "large" and "undefined" be removed?  Or should the goblin start out large?

iluwatar:
Ok @mattj256 you can start work on this. I think we can remove the unused enumerations.

dosdebug:
@iluwatar A Goblin initially starts with `NORMAL` size. Casting a shrink spell will make it `SMALL`. What if a shrink spell is casted again? Should size be cyclic e.g. `Small -> Normal -> Large -> Small`? Or, if Goblin is already `SMALL`, casting shrink spell should not work?

iluwatar:
@dosdebug I would say shrink spell makes smaller until the smallest possible (SMALL) after which it has no effect.

dosdebug:
@iluwatar Sure, a PR is submitted to you. Please see when you have time.

PR #641, opened 2017-09-21
Removes unused size and visibility enumerations

dosdebug:
This PR is related to #547 which seems not a question to me but removed unused enumerations to avoid misunderstanding.

### DIFF
--- a/command/src/main/java/com/iluwatar/command/Wizard.java
+++ b/command/src/main/java/com/iluwatar/command/Wizard.java
@@ -40,7 +40,9 @@ public class Wizard {
   private Deque<Command> undoStack = new LinkedList<>();
   private Deque<Command> redoStack = new LinkedList<>();
 
-  public Wizard() {}
+  public Wizard() {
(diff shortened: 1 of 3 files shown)


command/src/main/java/com/iluwatar/command/Wizard.java
```diff
@@ -40,7 +40,9 @@
   private Deque<Command> undoStack = new LinkedList<>();
   private Deque<Command> redoStack = new LinkedList<>();
 
-  public Wizard() {}
+  public Wizard() {
```
rastdeepanshu: Is there any use for keeping this constructor explicitly?

dosdebug: Not really but I wanted to keep the change as small as possible.

iluwatar: Yes, we should probably remove it.

iluwatar:
The changes look good. Thanks @dosdebug :+1:

PR #645, opened 2017-09-24
Removed redundant times(1) params for verify()

amitdash291:


rastdeepanshu:
👍

npathai:
@amitdash291 Thanks for contribution. It is merged. Though I feel that keeping `times(1)` makes things more explicit and readable.

amitdash291:
@npathai for readability it made sense to not remove it, though, on the other hand, most of the devs implicitly read `times(1)` when it isn't present

PR #646, opened 2017-09-25
Modifications for intermittent test failure in Throttling pattern.

rastdeepanshu:
Used AtomicLong instead of Integer in ConcurrentHashMap to make the increment calls atomic.
Ref: [643](https://github.com/iluwatar/java-design-patterns/issues/643)


throttling/src/main/java/com/iluwatar/throttling/CallsCount.java
```diff
@@ -22,51 +22,56 @@
...
 /**
  * A class to keep track of the counter of different Tenants
  * @author drastogi
  *
  */
 public final class CallsCount {
-  private static Map<String, Integer> tenantCallsCount = new ConcurrentHashMap<>();
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(CallsCount.class);
+  private static Map<String, AtomicLong> tenantCallsCount = new ConcurrentHashMap<>();
```
npathai: Question: Why change from int to long? Why not `AtomicInteger`? Is it because the call count can breach MAXINT limit?

throttling/src/test/java/com/iluwatar/throttling/B2BServiceTest.java
```diff
@@ -36,14 +36,15 @@
   @Test
   public void dummyCustomerApiTest() {
     Tenant tenant = new Tenant("testTenant", 2);
-    Throttler timer = new ThrottleTimerImpl(10);
+    Throttler timer = new ThrottleTimerImpl(100);
     B2BService service = new B2BService(timer);
-    
+
+    long counter = 0;
```
npathai: Can you explain briefly why are these changes needed?

throttling/src/test/java/com/iluwatar/throttling/B2BServiceTest.java
```diff
@@ -36,14 +36,15 @@
   @Test
   public void dummyCustomerApiTest() {
     Tenant tenant = new Tenant("testTenant", 2);
-    Throttler timer = new ThrottleTimerImpl(10);
+    Throttler timer = new ThrottleTimerImpl(100);
```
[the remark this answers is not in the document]
rastdeepanshu: Yes, sometimes the counter was being reset within the 10 ms period. To avoid that, the value is increased.

throttling/src/main/java/com/iluwatar/throttling/CallsCount.java
rastdeepanshu: Yes

throttling/src/test/java/com/iluwatar/throttling/B2BServiceTest.java
rastdeepanshu: To be honest, I am unable to figure out this. The code works perfectly this way but if the counter is calculated outside the for loop(like done previously), then sometimes it is not reflecting the proper changes.

rastdeepanshu: I am sorry. I think I tested this method before incrementing the count limit from 10 to 100(which was the root of intermittent failure). Will change this back to the original

rastdeepanshu:
@npathai you have my changes and comments.

PR #648, opened 2017-09-26
added missing whitespace to the elements of the unordered list

Peter-Morawski:


npathai:
@Peter-Morawski A unit test case of Balking pattern is causing the build to fail. Can you run build locally to verify?

Peter-Morawski:
@npathai I ran all test cases successfully and could even build and run the balking pattern without any problems.

image

image

image

npathai:
@iluwatar Any idea what might be causing this? Diff shows no change in tests, but still build has failing tests for Balking pattern

iluwatar:
@npathai could not reproduce. The fail is not related to this pull request anyway so I think this can be accepted. Please raise a new issue about the Balking pattern test failure so we can investigate further.

Issue #176, opened 2015-07-27
Wire Tap (EIP)


iluwatar:

JaXt0r:
Some Informations:
* Description from EIP: http://www.enterpriseintegrationpatterns.com/patterns/messaging/WireTap.html
* Implementation inside Apache Camel: http://camel.apache.org/wire-tap.html

codinghog:
I can help you out with most EIP patterns, starting from this one.

iluwatar:
Thanks @codinghog go ahead :+1:

PR #651, opened 2017-10-04
#176 Added new pattern - EIP Wire Tap

codinghog:
New pattern added - EIP Wire Tap.
- Added new directory - eip-wire-tap
- Added sources for wire tap pattern
- Added tests
- Added description to README.md file


codinghog:
I wanted to create fully working EIP context. `App.java` sets up and starts Spring Boot application, which takes care of adding all the routes to the context. I can refactor it by dropping Spring Boot, setting up Apache Camel context by hand, running routes and terminating the app.

However, test for `App.java` will test only if there were any exceptions thrown during the execution. Do you want any tests for the route? I have checked the Message Channel pattern and there are no tests for actual route, only for `App.java` execution.

iluwatar:
> However, test for App.java will test only if there were any exceptions thrown during the execution

Yes, this is fine. Not sufficient testing alone, but it is a good addition to test that all our examples run correctly.

> Do you want any tests for the route? I have checked the Message Channel pattern and there are no tests for actual route, only for App.java execution.

It is a big bonus if you can do that. I'm not that experienced with Camel and couldn't do that with Message Channel example.

codinghog:
Ok. I'll refactor `App.java` a little and add a test for it.

codinghog:
I have made changes according to your requirements. Unfortunately the build fails because of some test errors in `throttling` project. Any ideas?

iluwatar:
@codinghog is your branch up-to-date with master? A fix was made to Throttling pattern a while ago, see https://github.com/iluwatar/java-design-patterns/issues/643

codinghog:
@iluwatar All tests passed, changes made as you have requested

iluwatar:
Good job @codinghog! Thanks for the contribution :+1:

PR #652, opened 2017-10-07
Fix FactoryMethodTest and PrototypeTest

yosfik:
- The test message for the WeaponType assertion should tell the expected WeaponType instead of the Weapon class name.
- Remove compiler warning of unused generic type in PrototypeTest


iluwatar:
Thanks @yosfik :+1:

PR #654, opened 2017-10-16
maybe it should be "Let's" here

fanofxiaofeng:


iluwatar:
Yes, you are right @fanofxiaofeng

PR #656, opened 2017-10-20
Use UserConverter instead of rewriting its ctor

bjurga:
UserConverter() is never used.


iluwatar:
Looks good @bjurga :+1:

Issue #653, opened 2017-10-08
Searching/Filtering  at http://java-design-patterns.com/patterns/ doesn't work


tomyy:
* I tried to search for "abstract": all patterns a shown, no indication that anything was searched
* I tried to restrict the categories to 'Creational': still all patterns are shown
* I tried to restrict the tags to 'Gang of Four': still all patterns are shown

To make sure that it is not my browser I tried to do this with Mozilla Firefox 56.0, Google Chrome 61.0.3163.100 and Microsoft Edge 40.15063.0.0 and I had the same problem with every one of these browsers.

aldaircruz:
I have the same issue, I got those errors:

![imagem](https://user-images.githubusercontent.com/17972991/31317532-78664386-ac3a-11e7-8ae5-6a51c900aedb.png)

![imagem](https://user-images.githubusercontent.com/17972991/31317534-96ddaff2-ac3a-11e7-9036-f134ab1ba54e.png)

iluwatar:
Thanks for reporting @tomyy and @aldaircruz We are looking into it.

dungnart:
I can not do any Search/Filter on http://java-design-patterns.com/patterns/ site. Please check it. Thank you !

nathancyam:
It seems if the patterns' content contains any braces, it breaks JSON parsing. I've got a PR here (https://github.com/iluwatar/java-design-patterns/pull/657) that adds the `url_encode` filter to ensure this doesn't happen again.

The build seems to be breaking, but it looks like Travis can't resolve certain URLs.

PR #657, opened 2017-10-22
#653: Url encode content to fix JSON breaking pattern descriptions

nathancyam:
Some of the patterns contain braces that will break jQuery's
`getJSON` method when it attempts to parse the response. Adding
the `url_encode` filter ensures that this doesn't happen again.


iluwatar:
Well done @nathancyam This contribution is valuable :+1:

PR #665, opened 2017-11-08
(Abstract Factory) Add factory of factories

Tschis:
The explanation for the pattern mentions that it is "a factory of factories", but the example does not have any factory class that returns the concrete classes of **KingdomFactory**.

This commit removes the direct instantiation of **ElfKingdomFactory** and **OrcKingdomFactory** from **App.main()** and delegates their creation to a new (factory of factories) **FactoryMaker** inner class of **App**.

Even though the pattern itself does not necessarily need this, it helps to synchronize the explanation text with the coding example, making it easier to understand that statement.

### DIFF
--- a/abstract-factory/README.md
+++ b/abstract-factory/README.md
@@ -120,6 +120,45 @@ king.getDescription(); // Output: This is the Elven king!
 army.getDescription(); // Output: This is the Elven Army!
 ```
 
+Now, we can design a factory for our different kingdom factories. In this example, we created FactoryMaker, responsible for returning an instance of either ElfKingdomFactory or OrcKingdomFactory.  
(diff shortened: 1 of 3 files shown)


Tschis:
Sorry, I had problems with the checkstyle on the first commit. I have since configured it properly and made changes (3634b33) to my code so that it would follow the style guidelines.

After that I had troubles with the correct command for running the maven checkstyle, so I needed an extra commit (933c84f). 

My code is now running locally with build success on 
`install -DskipTests=true -Dmaven.javadoc.skip=true -B -V -e`

Tschis:
Done, @iluwatar

iluwatar:
Thank you @Tschis for this improvement, much appreciated :+1:

PR #668, opened 2017-11-19
added synchronized keyword to method that reads from server poll

### LINKED ISSUE #666: Is it threadsafe ?
Seems like lastServedId field from LoadBlanacer.class is shared among all instances.
I think this is not threadsafe and will crash when mutiple threads will try to increment it.

https://github.com/iluwatar/java-design-patterns/blob/390c33e36e82b558e8cab6028d96e49c0a20da9f/monostate/src/main/java/com/iluwatar/monostate/LoadBalancer.java#L39

Krandelbord:
fixes #666 to make it threadsafe
BTW: Moved id to static initizalizer block since it used only here.


iluwatar:
Thanks @Krandelbord the fix looks good :+1:

PR #676, opened 2017-11-30
Add presentation for Abstract Factory

Tschis:
Following work of #197, adding presentation for Abstract Factory pattern.

@iluwatar I'm not sure how the licenses work, so I put my name on MIT license in the presentation.html file. Since I based mine on your Hexagonal presentation, if that should be your name in there let me know so I can change it.


iluwatar:
Thanks @Tschis looks good :+1:

PR #678, opened 2017-12-11
Add presentation for Command and Factory Method

Tschis:
Following work of #197, adding presentations for Command and Factory Method.


iluwatar:
Looks great @Tschis keep 'em coming :+1:

PR #680, opened 2017-12-14
replace 'a' with 'an'

fanofxiaofeng:
[A or An](https://www.a-or-an.com/a_an/axe)


iluwatar:
Thanks @fanofxiaofeng :+1:

PR #687, opened 2017-12-29
handleRequest method in RequestHandler can introduce 'template method'

maxtmn:
with hook method. Also Request.isHandeld() can be used to interrupt
chain

Pull request title

- Clearly and concisely describes what it does
- Refer to the issue that it solves, if available

Pull request description


chain/src/main/java/com/iluwatar/chain/RequestHandler.java
```diff
@@ -44,12 +44,15 @@ public RequestHandler(RequestHandler next) {
    * Request handler
    */
   public void handleRequest(Request req) {
+    if (!req.isHandled()) {
+      handleHook(req);
+    }
```
iluwatar: This causes `xxx handling request xxx` to be printed although they are not handling it

maxtmn: right, it can be so, simplifyied even more. need to pull, than, request type as an input parameter into constructor

iluwatar:
Pull request closed due to inactivity

Issue #694, opened 2018-01-05
Remove redundant explicit super() constructor calls


lotsabackscatter:
The super() calls (taking no arguments) are inserted during compile time, so just make the code more verbose than required.

One example is found in the [HeroStat ](https://github.com/iluwatar/java-design-patterns/blob/master/value-object/src/main/java/com/iluwatar/value/object/HeroStat.java#L40) class.

james-a-roland:
@iluwatar I'll take this if no one else has

iluwatar:
@james-a-roland there is a pull request about this, see https://github.com/iluwatar/java-design-patterns/pull/695. But I have a feeling it does not solve all redundant calls to `super()`. Either work with @jonatasemidio by reviewing the PR or create your own PR that resolves all the cases. I will assign the issue to you.

jonatasemidio:
Hi @james-a-roland and @iluwatar,
Do you know what type of unit test can I include to check this other cases?

iluwatar:
I would probably just `grep` through all `super` occurrences

james-a-roland:
@iluwatar - Pull request can be found here https://github.com/iluwatar/java-design-patterns/pull/697

jonatasemidio:
@iluwatar, another consideration... I know that grep can find all occurrences, but we might have some samples that super() calls some father's constructors.

Hi @james-a-roland, To solve #694 ... How did you checked all super() occurrences?

olcbean:
I used `grep 'super()'`

james-a-roland:
Same as @olcbean - any calls to super() called from a custom constructor were deleted. Since all objects are POJO's, I think this is a safe operation.

jonatasemidio:
So... In this case, we can use #697 instead of #695

PR #695, opened 2018-01-05
 Remove redundant explicit super()

jonatasemidio:
Remove redundant explicit super() constructor calls #694

Remove redundant explicit super() constructor

The super() calls (taking no arguments) are inserted during compile time, so just make the code more verbose than required.

### DIFF
--- a/value-object/src/main/java/com/iluwatar/value/object/HeroStat.java
+++ b/value-object/src/main/java/com/iluwatar/value/object/HeroStat.java
@@ -37,7 +37,6 @@ public class HeroStat {
 
   // All constructors must be private.
   private HeroStat(int strength, int intelligence, int luck) {
-    super();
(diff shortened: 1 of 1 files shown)


olcbean:
I think there are a few more occurrences of `super()`. @jonatasemidio do you have time to remove them too?

jonatasemidio:
Hi @olcbean is totally fine, but now I need to check if it is necessary because @james-a-roland created another and more complete PR (#697).

@iluwatar, another consideration... I know that grep can find all occurrences, but we might have some samples that super() calls some father's constructors. 

Hi @james-a-roland, How did you checked all super() occurrences?

jonatasemidio:
Sorry @iluwatar and @james-a-roland... You don't need to answer here because I've already add the last comment in the right discussion #694

PR #698, opened 2018-01-15
Remove redundant explicit super() constructor calls #694

olcbean:
This PR removes few more explicit calls to `super()`.

Relates to #694


iluwatar:
Thanks @olcbean :+1:

Issue #701, opened 2018-01-20
a new design pattern : efficient-cache


zhengzhanpeng:
I have an idea for building an efficient cache design to solve the problem of needing to calculate the same situation frequently in some scenarios. I'm working on it. I am annotating now, the function has been achieved.I hope I can get your approval.

PR #705, opened 2018-01-24
#701 new pattern: efficient_cache_result

zhengzhanpeng:
efficient_cache_result

- Efficient cache result is a high-performance pattern for saving cached results under concurrency conditions

Pull request description

- I want to try to make this model. If there are problems in the code, please point out.


berezowskil:
Mute

24.01.2018 1:19 PM "Albert" <notifications@github.com> napisał(a):

PR #706, opened 2018-01-24
#701 new pattern: efficient_cache_result

zhengzhanpeng:
efficient_cache_result

- Efficient cache result is a high-performance pattern for saving cached results under concurrency conditions

Pull request description

- I want to try to make this model. If there are problems in the code, please point out.


zhengzhanpeng:
@iluwatar If you have time, I would like to git your suggestion.

iluwatar:
I don't see any changes. Where is the code?

zhengzhanpeng:
image
Here it is.

zhengzhanpeng:
Can't you see that? @iluwatar

zhengzhanpeng:
Sorry, I git.  I  will solve immediately it in two minutes

Issue #508, opened 2016-10-20
SonarQube reports code smells


iluwatar:
Java Design Patterns is analyzed with SonarQube.com static analysis. The analysis shows some blocker level code smells that should be fixed:

https://sonarqube.com/component_issues/index?id=com.iluwatar%3Ajava-design-patterns#resolved=false|severities=BLOCKER|types=CODE_SMELL

andreybleme:
@iluwatar Hello. I can help with this one :100: 
I'm on it.

sangupta:
I will take care of this.

iluwatar:
Thank you @sangupta you're good to go!

iluwatar:
@sangupta are you still working on this?

sangupta:
@iluwatar Am sorry - have been pressed for time lately. Will have sometime in March - when I should be able to lend a helping hand.

Praveer-grover:
@iluwatar I can begin on this if no one else is looking at it, to start with I was just going through delegation and builder pattern. I can create a pull request with sonar fixes for those, is it okay?

iluwatar:
@Praveer-grover this is ok, please go ahead.

mookkiah:
Updated sonarqube url
https://sonarcloud.io/project/issues?id=com.iluwatar%3Ajava-design-patterns&resolved=false&severities=BLOCKER&types=CODE_SMELL

@Praveer-grover Are you still working on these issues? Let me know if you want me to share some modules?

iluwatar:
Ping @Praveer-grover

dosdebug:
@mookkiah Are you planning to completely work on this ticket or you need a helping hand? ;)

mookkiah:
Either way is fine for me. As @Praveer-grover is not responding, I will look at sonarqube reported issues in prototype module (maybe this weekend). @dosdebug, you are welcome to help another module. Please just comment so we don't step on each other.

dosdebug:
@mookkiah I will be working on `dao` and `naked-objects-dom` then.

iluwatar:
Hey @mookkiah and @dosdebug are you guys still working on this?

mookkiah:
I will work on prototype module.

dosdebug:
@iluwatar No Sir, not working on this.

olcbean:
Note : 
[...]

PR #709, opened 2018-01-27
issue 508 - using copy constructor to implement prototype.

mookkiah:
As per http://www.artima.com/intv/bloch13.html, clone method in java is
broken or confusing. So staying away from clone method.

SonarQube reports code smells #508

- SonarQube recommends Remove this "clone" implementation; use a copy constructor or copy factory instead

Pull request description

- introduced copy method which uses copy constructor. keeps the implementation transparent and expect developer to make deep copy during copy operation.


iluwatar:
Thanks @mookkiah for this fix! :+1:

PR #713, opened 2018-02-07
Use the @TestInstance annotation

baislsl:
Use the TestInstance annotation in class to avoid ```static``` and  enable inhertied of methods refered by MethodSource.


iluwatar:
Thanks @baislsl for the improvement :+1:

PR #723, opened 2018-03-07
Captain is implementing RowingBoat - Adapter pattern

Juaanma:
Captain and RowingBoat aren't interchangable. 
Captain, as a client, may have a method to use the boat, but it isn't needed that they share the same interface, as it makes the code more confusing.


Juaanma:
It should be working now.

iluwatar:
Thanks @Juaanma for this improvement :+1:

Issue #602, opened 2017-07-25
Simplify Multiton implementation


iluwatar:
The implementation could be simplified as commented here: http://java-design-patterns.com/patterns/multiton/#comment-3432626979

gopinath-langote:
@iluwatar  - I am interested in same. Is it ok If I took the different example that the example you have its code?

iluwatar:
Ok @gopinath-langote

iluwatar:
@gopinath-langote  are you still implementing this?

gopinath-langote:
@iluwatar : Give me some time. Sorry, couldn't get time to work on this. Please reopen this issue for enhancement so that someone else can also work on this.

Thanks.

iluwatar:
This issue is free for taking again

anthonycampbell:
@iluwatar  - I am working on it.

iluwatar:
Ok @anthonycampbell

anthonycampbell:
Hey, I am having trouble. I do not really understand what "Creating an enum (e.g. Nazgul) with the getName method" means. I see the enum in NazgulName. Is this person talking about a different enum? I am wondering because they say "e.g Nazgul" rather than NazgulName. I am not sure I see how putting an Enum in getName() would simplify things. If there is a better place for me to ask this questions please tell me. I just really want to contribute to this project.

iluwatar:
My understanding is that `Nazgul` should be implemented as enum instead of class to make the example simpler.

anthonycampbell:
I might understand better now. Like basically replace the word class in Nazgul.java with the word enum, and then fix the errors, and refactor so that it all still works. Is that right?

iluwatar:
You are on the right track. We are trying to create enum based Multiton (see enum based Singleton here https://github.com/iluwatar/java-design-patterns/blob/master/singleton/src/main/java/com/iluwatar/singleton/EnumIvoryTower.java). The method `getName` is not important, it's only there for the test.

anthonycampbell:
I think I got it now. Seems very simple. I will post here again if I have more trouble for some reason. Thank you.

PR #724, opened 2018-03-09
My solution to issue #602. 

anthonycampbell:
Implements multiton with enum in file called NazgulEnum.java. Adds the enum example to App.java. Adds a test for the enum called NazgulEnumTest.java. If you could please let me know your thoughts then I would much appreciate that.


iluwatar:
Looks good @anthonycampbell, thanks! :+1:

PR #735, opened 2018-04-06
Trampoline pattern logger fix

semiheker89:
Trampoline pattern has logger object build error.
The error fixed with using **org.slf4j.Logger**.

Instead of using anatation, Logger object is created for logging.


iluwatar:
@semiheker89 this pull request is about to expire. Are you still working on it?

iluwatar:
Closed due to inactivity

PR #740, opened 2018-04-09
Correct the command for computing code coverage

mernst:
The command in file `CODE_COVERAGE.md` creates `jacoco.exec` files but not the reports.
This pull request changes the command to generate the reports, making the rest of the file correct.


iluwatar:
Thanks @mernst well spotted!

PR #748, opened 2018-04-28
Parameters in Converter class methods was renamed. 

pelmegov:
Parameters in Converter class methods was renamed. 
A base class must not contain the concrete entities names or concrete class names


iluwatar:
Many thanks @pelmegov for this improvement! :+1:

PR #752, opened 2018-05-26
#643 Fix flaky tests for throttling pattern

christophercolumbusdog:
I took a look at the Travis logs and the troublesome test. From examining the code and tinkering around, my best guess is that in the Travis environment, the threat that resets the API access limit may prematurely reset the count, making the assertion fail.

I have instead changed the throttler to be an empty implementation (to avoid any unintended resets, which is not the point of the test). Additionally, I changed it to an assertEquals so that in the event of another failure, we can see exactly what the counter ended up being, which will help with debugging.

### DIFF
--- a/throttling/src/test/java/com/iluwatar/throttling/B2BServiceTest.java
+++ b/throttling/src/test/java/com/iluwatar/throttling/B2BServiceTest.java
@@ -22,29 +22,27 @@
  */
 package com.iluwatar.throttling;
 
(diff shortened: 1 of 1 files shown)


iluwatar:
Thanks @christophercolumbusdog for looking into this. Let's merge the fix candidate and see how it behaves.

npathai:
@christophercolumbusdog Thanks for the fix :+1:

PR #774, opened 2018-07-18
Add an 'Override' annotaion

zenuo:
Add an annotation 'Override' to mothod 'AppClient.TcpLoggingClient#run'.


iluwatar:
Thanks @zenuo :+1:

Issue #781, opened 2018-08-10
ClassCastException was throwed in App.java#52


yushouling:
The row that has an exception was:
hayes.accept(conUnix); // Hayes modem with Unix configurator

Because of ConfigureForUnixVisitor didn't implements HayesVisitor.

#java-design-patterns/acyclic-visitor

npathai:
@iluwatar That is not a bug, it is expected to happen. I will change the log level and use `instanceof` instead to resolve the confusion.

iluwatar:
Well done @npathai Thank you!

PR #783, opened 2018-08-13
#781: Resolve ClassCastException in App.java 

SauravP97:
#781 Resolve ClassCastException in App.java 
      Resolved ClassCastException thrown because ConfigureForUnixVisitor didn't implements HayesVisitor. 
Implemented AllModemVisitor in ConfigureForUnixVisitor class and added visit(Hayes hayes) method.

### DIFF
--- a/acyclic-visitor/src/main/java/com/iluwatar/acyclicvisitor/ConfigureForUnixVisitor.java
+++ b/acyclic-visitor/src/main/java/com/iluwatar/acyclicvisitor/ConfigureForUnixVisitor.java
@@ -26,13 +26,17 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /**
- * ConfigureForUnixVisitor class implements zoom's visit method for Unix 
+ * ConfigureForUnixVisitor class implements zoom's and hayes' visit method for Unix 
  * manufacturer
  */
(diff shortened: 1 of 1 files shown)


SauravP97:
What just happened !! ... Why the Travis CI build failed . Can anyone explain me the possible reasons

npathai:
@SauravP97 Actually the `ClassCastException` is expected. The whole point of Acyclic visitor is that you can only implement the type interfaces that you are actually interested in visiting. In this case `ConfigureForUnixVisitor` is only interested in visiting `Zoom` modems. So when called with `Hayes` modem, it just skips calling visit on that visitor because types don't match.

I agree that classcast is not a proper way, I will change it to using `instanceof` check instead and remove the error log to info log with proper message. Thanks for the PR though. Let me know if you have any futher doubts.

SauravP97:
Thanks a lot for clearing that issue. Actually this is my first PR to a repository so I'm a kind of newbie.
 So can I ask whether my PR will get merged or not ?

npathai:
@SauravP97 I apologize that I won't be able to accept this PR in current state because it doesn't solve the actual issue. But I surely can guide you solving it the proper way if you would still like to contribute on this issue. Let me know.

npathai:
@SauravP97 If you are interested then reply. Else we can close this PR.

npathai:
Closing this in favor of PR #794

PR #785, opened 2018-08-20
Fix converter image Display Issue

sgyyz:
In converter pattern, README.md has a wrong image reference due to url is case-sensitive.


npathai:
@sgyyz Thanks for the PR :+1:

PR #791, opened 2018-09-06
Execute Around - use lambda

er2:
Refactor the Execute Around pattern to use a lambda. Lambdas make such patterns feel lighter weight.

I first heard about this pattern in a talk by Venkat Subramanian. The thrust of the talk was that lambdas made such patterns nicer.

I extracted the lambda to its own variable to make clearer the type it's implementing.

### DIFF
--- a/execute-around/src/main/java/com/iluwatar/execute/around/App.java
+++ b/execute-around/src/main/java/com/iluwatar/execute/around/App.java
@@ -22,7 +22,6 @@
  */
 package com.iluwatar.execute.around;
 
-import java.io.FileWriter;
 import java.io.IOException;
 
 /**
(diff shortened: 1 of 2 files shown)


execute-around/src/main/java/com/iluwatar/execute/around/App.java
```diff
@@ -42,14 +41,11 @@
...
-    new SimpleFileWriter("testfile.txt", new FileWriterAction() {
-
-      @Override
-      public void writeFile(FileWriter writer) throws IOException {
-        writer.write("Hello");
-        writer.append(" ");
-        writer.append("there!");
-      }
-    });
+    FileWriterAction writeHello = writer -> {
```
IAmPramod: We should add `@FunctionalInterface` annotation to `FileWriterAction`.

er2: Done. I did a fix, soft reset, force push since I wanted to fix my email address in the last commit.

npathai:
@er2 Seems good to go. Thanks for the PR.

PR #793, opened 2018-09-11
Fix redundant list alloc in LotteryNumbers

andrievsky:
Fix redundant list alloc in LotteryNumbers

### DIFF
--- a/hexagonal/src/main/java/com/iluwatar/hexagonal/domain/LotteryNumbers.java
+++ b/hexagonal/src/main/java/com/iluwatar/hexagonal/domain/LotteryNumbers.java
@@ -22,13 +22,12 @@
  */
 package com.iluwatar.hexagonal.domain;
 
-import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashSet;
-import java.util.List;
 import java.util.PrimitiveIterator;
 import java.util.Random;
 import java.util.Set;
+import java.util.Iterator;
 
 /**
  *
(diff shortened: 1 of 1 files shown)


hexagonal/src/main/java/com/iluwatar/hexagonal/domain/LotteryNumbers.java
```diff
@@ -85,11 +84,10 @@ public static LotteryNumbers create(Set<Integer> givenNumbers) {
    * @return numbers as comma separated string
    */
   public String getNumbersAsString() {
-    List<Integer> list = new ArrayList<>();
-    list.addAll(numbers);
     StringBuilder builder = new StringBuilder();
+    Iterator<Integer> iterator = numbers.iterator();
     for (int i = 0; i < NUM_NUMBERS; i++) {
-      builder.append(list.get(i));
+      builder.append(iterator.next());
```
IAmPramod: Why we don't have a check for `iterator.hasNext()`

andrievsky: The same reason we didn't check a list boundaries.

hexagonal/src/main/java/com/iluwatar/hexagonal/domain/LotteryNumbers.java
```diff
@@ -85,11 +84,10 @@ public static LotteryNumbers create(Set<Integer> givenNumbers) {
    * @return numbers as comma separated string
    */
   public String getNumbersAsString() {
-    List<Integer> list = new ArrayList<>();
-    list.addAll(numbers);
     StringBuilder builder = new StringBuilder();
+    Iterator<Integer> iterator = numbers.iterator();
```
npathai: Can't we directly use foreach loop and not depend on `NUM_NUMBERS`?

iluwatar: I think so too. However, I'll accept this PR as an improvement and a step into the right direction.

iluwatar:
Thanks @andrievsky :+1:

Issue #775, opened 2018-07-18
Implement Retry pattern with exponential backoff 


awesomefei:
In practice, we often retry with exponential backoff to avoid making traffic congestion/server load worse, would be nice to implement this option.

Also, I would like to implement this.

iluwatar:
Yes, I agree. Go ahead @awesomefei

AnaghaSasikumar:
Hi, is this done?

AnaghaSasikumar:
Created pull request #797 in case it's still needed. I am new to open source, would really appreciate any feedback. I used this source to do it: https://cloud.google.com/storage/docs/exponential-backoff

Issue #562, opened 2017-04-01
Spatial Partition pattern


iluwatar:
http://gameprogrammingpatterns.com/spatial-partition.html

szoeke:
hi @iluwatar 
I'm willing to implement this pattern.

iluwatar:
Ok @szoeke

prakis:
@szoeke This is pretty interesting pattern, please let me know if you need help in implementing/discussing this pattern.

iluwatar:
This issue is free for taking again

danielsiwiec:
@iluwatar I'll take a stab at this

iluwatar:
Ok @danielsiwiec

iluwatar:
@danielsiwiec  are you still implementing this?

danielsiwiec:
@iluwatar sorry, no - I got side tracked into other things. Please take me off of it for the time being.

iluwatar:
This issue is free for taking again

karmagamer:
I'd like to try to implement this. This looks interesting. @iluwatar

iluwatar:
Ok @karmagamer

karmagamer:
@iluwatar I am on it. I will update you once I am finished,

tliecau:
@karmagamer  please let me know if you need help with implementing or something with this pattern. I would love to help

karmagamer:
@tliecau Sure buddy, so far it's coming along fine. I will update you in a week if I am stuck or need help.

AnaghaSasikumar:
Hi, is this done? 
I have been working on something for the last few days and it is working...if it's not done can I submit a PR?

AnaghaSasikumar:
Ok sure @karmagamer i'll give it a shot then 👍😀

AnaghaSasikumar:
PR #797

PR #797, opened 2018-09-29
Retry pattern with exponential back-off, Spatial Partition Pattern #775 #562

AnaghaSasikumar:
Added new implementation of retry pattern with the exponential back-off : delay time increases exponentially with number of attempts, stops increasing once maxDelay is reached.
Update: Added Spatial Partition design pattern.


retry/src/main/java/com/iluwatar/retry/RetryWithExponentialBackoff.java
```diff
@@ -0,0 +1,117 @@
...
+                return this.op.perform();
+            } catch (BusinessException e) {
+                this.errors.add(e);
+
+                if (this.attempts.incrementAndGet() >= this.maxAttempts || !this.test.test(e)) {
+                    throw e;
+                }
+
+                try {
+                    long testDelay = Math.pow(2, this.attempts)*1000 + Math.round(Math.random()*1000);
```
IAmPramod: I don't think using `Math.random()` will give increasing backoff. We can think of using fibonacci series for delay.
Something similar to https://carlosbecker.com/posts/exponential-backoff-java8/

AnaghaSasikumar: Hi, thanks for responding! :D
Actually, I was trying to do it the way it was suggested in the link, they basically added a random number to 2^number_of_attempts. I think Fibonacci should work too though. Should it be changed to that?

IAmPramod: @AnaghaSasikumar adding random number to 2^no_of_attempts doesn't guarantee that back-off is increasing in order.

AnaghaSasikumar: That part will be < 1s, Math.random() generates a random double between 0 and 1.

coryallen34:
I don't want nothing from you you f******

AnaghaSasikumar:
I don't understand why the build is failing, it is working well on eclipse in my system. But eclipse is using Java SE-10, is that the reason? Could you tell me how else I should be building it, because I think the code could be useful...the documentation I can work on if you decide to use it.

iluwatar:
@AnaghaSasikumar the build is failing due to PMD violation.

```
[INFO] --- maven-pmd-plugin:3.6:check (default) @ retry ---
[INFO] PMD Failure: com.iluwatar.retry.RetryWithExponentialBackoff:105 Rule:UselessParentheses Priority:4 Useless parentheses..
```

npathai:
@iluwatar There are too many commits, don't you think we should squash em into a single commit?

IAmPramod:
@npathai @iluwatar According to me, we should squash commits while merging a PR. This will help us keep track of commits related to a particular issue.
thoughts?

AnaghaSasikumar:
Is there something I can do?
I took some time to get a hang of it all, I should do better next time...do let me know if I can do something for this though

iluwatar:
@IAmPramod @npathai yes, I think we can start squashing every pull request as they come in. I can change it in the repository's settings to allow only squash merges.

retry/.checkstyle
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<fileset-config file-format-version="1.2.0" simple-config="true" sync-formatter="false">
+  <fileset name="all" enabled="true" check-config-name="java-design-pattern-repo" local="false">
+    <file-match-pattern match-pattern="." include-pattern="true"/>
+  </fileset>
+</fileset-config>
```
iluwatar: You should not need these local Checkstyle configurations (`.checkstyle`). Please remove. Check elsewhere too.

spatial-partition/README.md
```diff
@@ -0,0 +1,27 @@
...
+</p><p>
+In our implementation, we use the quadtree data structure which will reduce the time complexity of finding the objects within a certain range from O(n^2) to O(nlogn), increasing the speed of computations immensely in case of large number of objects.
+ </p>
+
+<h2>Tutorials (Credits)</h2>
+
+<p><a href = "http://gameprogrammingpatterns.com/spatial-partition.html">Spatial Partition (Game Programming patter/Optimisation pattern)</a></p>
+<p><a href = "https://www.youtube.com/watch?v=OJxEcs0w_kE">Quadtree tutorial Part 1</a></p>
+<p><a href = "https://www.youtube.com/watch?v=QQx_NmCIuCY">Quadtree tutorial Part 2</a></p>
+<p><a href = "https://www.youtube.com/watch?v=z0YFFg_nBjw">Quadtree tutorial Part 3</a></p>
```
iluwatar: Please try to stick with the standard pattern template. The sections are listed here: https://github.com/iluwatar/java-design-patterns/wiki/02.-Pattern-template

spatial-partition/pom.xml
```diff
@@ -0,0 +1,20 @@
...
+  <artifactId>spatial-partition</artifactId>
+  <build>
+    <plugins>
+      <plugin>
+        <artifactId>maven-compiler-plugin</artifactId>
+        <version>3.7.0</version>
+        <configuration>
+          <release>10</release>
+        </configuration>
+      </plugin>
```
iluwatar: The `maven-compiler-plugin` is already defined in the main `pom.xml`. You should not define your own version here.

spatial-partition/src/main/java/com/iluwatar/spatialpartition/Bubble.java
```diff
@@ -0,0 +1,79 @@
...
+
+public class Bubble extends Point<Bubble> {
+
+  int radius;
+
+  Bubble(int x, int y, int id, int radius) {
+    this.x = x;
+    this.y = y;
+    this.id = id;
+    this.radius = radius;
```
iluwatar: These fields could be `final`

spatial-partition/src/test/java/com/iluwatar/spatialpartition/BubbleTest.java
```diff
@@ -0,0 +1,88 @@
...
+
+class BubbleTest {
+
+  @Test
+  void test() {
+    moveTest();
+    touchesTest();
+    popTest();
+    handleCollisionTest();
+  }
```
iluwatar: Move them to individual tests.

spatial-partition/src/test/java/com/iluwatar/spatialpartition/QuadTreeTest.java
```diff
@@ -0,0 +1,77 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package com.iluwatar.spatialpartition;
+
+/**
+ * Testing QuadTree class.
+ */
```
iluwatar: Put the comment directly above the class. Check elsewhere too.

spatial-partition/src/test/java/com/iluwatar/spatialpartition/QuadTreeTest.java
```diff
@@ -0,0 +1,77 @@
...
+import static org.junit.jupiter.api.Assertions.*;
+import java.util.ArrayList;
+import java.util.Hashtable;
+import java.util.Random;
+import org.junit.jupiter.api.Test;
+
+class QuadTreeTest {
+
+  @Test
+  void test() {
```
iluwatar: The test method should have a descriptive name, not a generic one.

iluwatar: Check this elsewhere too.

iluwatar:
@AnaghaSasikumar I posted the review comments. Please comment when you have implemented the changes. Thanks!

iluwatar:
Just to let you know it's not recommended to bundle multiple issues in the same PR as has been done here.

spatial-partition/src/main/java/com/iluwatar/spatialpartition/Bubble.java
AnaghaSasikumar: The x and y fields change when move is invoked, but I made the id and radius fields final.

AnaghaSasikumar:
@iluwatar I made changes according to your comments. Regarding the 2 issues in 1 PR issue I actually just made changes directly to master branch for both, I didn't realise it would automatically pull those commits too. Will create separate branches from now on. Please let me know in case there are other things to be fixed...at the risk of sounding repetitive, I am new to this so if you think there is something that should be fixed even in the coding style, I could make those changes too. Thanks!

AnaghaSasikumar:
I was looking this up and I think the problem is that I built it as a Maven project instead of a module...should I just delete this project and make a new module for the parent project? Or should I add it manually into the main pom, will that be okay, or will it cause some other errors?

iluwatar:
Yes, you can just add it manually to the main `pom.xml`. It is just a one-liner.

AnaghaSasikumar:
Maven clean I think has worked for me but the travis build is failing here now

iluwatar:
The pull request bundles multiple issues and cannot be accepted as-is. In your fork, please create a new branch from the `master` and commit there only the changes related to one issue e.g. retry pattern. Then create a new pull request.

PR #800, opened 2018-10-15
Fix small points in Strategy and Decorator pattern.

trumvekhuya:
- Remove redundant Exception from throws list in unit tests.
- If we assign new clubberTroll instance with the same variable, it will be wrapped by itself.


iluwatar:
Thanks @IAmPramod

iluwatar:
Thank you @trumvekhuya :+1:

PR #803, opened 2018-10-15
Resolves #643, test cases failed due to global state in CallsCount

### LINKED ISSUE #643: Intermittent test failure in Throttling pattern
See https://travis-ci.org/iluwatar/java-design-patterns/builds/278878833

npathai:
Resolves #643 

Root Cause
---
Test cases failed due to presence of global state in `CallsCount`. Because `AppTest` was executed before `B2BServiceTest`, it scheduled 1 sec timer using `ThrottleTimerImpl` class. While resetting it used that global `CallCount.reset()` method, which reset all counters. So that causes thread safety issue because of unintended sharing of application state between test cases, which is not a good practice.

Solution
---

Solution was not to have global state using `static` map. Rather global state should be maintained using single object and passing that as dependency wherever needed.

Todos
---
- [x] Update class diagram. (Once solution is approved)


iluwatar:
The proposed solution looks good

npathai:
@iluwatar Updated the class diagram. Ready to merge!

PR #805, opened 2018-10-15
Refactoring changes in Ambassador Pattern

npathai:
Changes
---
  - Updated test cases to use Junit Assert method as compared to assert keyword 
 - Proper testing of `RemoteService` using `RandomProvider` interface. Introduced `RandomProvider` interface so that randomness can be controlled from test cases. 
 - For readability used constant for representing FAILURE

Todos
---

- [x] Update class diagram once refactoring is approved -> no class diagram


ambassador/src/test/java/com/iluwatar/ambassador/ClientTest.java
```diff
@@ -35,6 +37,6 @@ public void test() {
     Client client = new Client();
     long result = client.useService(10);
 
-    assert result == 100 || result == -1;
+    assertTrue(result == 100 || result == -1);
```
IAmPramod: We should use `RemoteServiceInterface.FAILURE` here as well. Also, we should create a contstant for 100 as well and use int respective tests.

ambassador/src/test/java/com/iluwatar/ambassador/RemoteServiceTest.java
```diff
@@ -22,16 +22,43 @@
...
+    long result = remoteService.doRemoteFunction(10);
+    assertEquals(RemoteServiceInterface.FAILURE, result);
+  }
+
+  @Test
+  public void testSuccessfulCall() {
+    RemoteService remoteService = new RemoteService(
+        new StaticRandomProvider(0.2));
+    long result = remoteService.doRemoteFunction(10);
+    assertEquals(100, result);
```
IAmPramod: We should create a constant for 100.

ambassador/src/test/java/com/iluwatar/ambassador/ServiceAmbassadorTest.java
```diff
@@ -32,6 +34,6 @@
   @Test
   public void test() {
     long result = new ServiceAmbassador().doRemoteFunction(10);
-    assert result == 100 || result == -1;
+    assertTrue(result == 100 || result == RemoteServiceInterface.FAILURE);
```
IAmPramod: We should create a constant for 100.

ambassador.patch
```diff
@@ -0,0 +1,205 @@
+diff --git a/ambassador/src/main/java/com/iluwatar/ambassador/RemoteService.java b/ambassador/src/main/java/com/iluwatar/ambassador/RemoteService.java
```
IAmPramod: I think this is an unintentional file added to commit. We should remove this.

npathai: My bad. Will do that. Didn't realize. Thanks.

ambassador/src/test/java/com/iluwatar/ambassador/ClientTest.java
npathai: Oh yeah missed here. Done!
100 is calculated at runtime, so I guess we don't really need a constant as such.

iluwatar:
The solution looks good to me as well

npathai:
@iluwatar It doesn't have a class diagram :thinking: So we can merge this PR and I will add a new PR with class diagram if needed.

PR #806, opened 2018-10-16
Refactored Event Queue

npathai:
1) Test cases were not stopping AudioService before ending test case, but test comments suggested differently. 
```
// test that service is finished
 assertFalse(!audio.isServiceRunning()); // which is equivalent to service is running
```
2) Changed `Audio` to be a good singleton, previously because of `Audio` being bad singleton, test cases which were using static methods could have caused intermittent failures. After changes now each test case has dedicated instance of `Audio` service which provides isolation between test cases.

3) Made `stop()` method blocking till service stops

4) Made some other refactorings as well

Todo
---
- [x] Update class diagram once PR is approved.


event-queue/src/main/java/com/iluwatar/event/queue/Audio.java
```diff
@@ -38,49 +41,56 @@
  *
  */
 public class Audio {
+  private static final Logger LOGGER = LoggerFactory.getLogger(Audio.class);
+  private static final Audio INSTANCE = new Audio();
```
IAmPramod: We should use lazy initialization for creating instance of `Audio` class

event-queue/src/test/java/com/iluwatar/event/queue/AudioTest.java
```diff
@@ -39,6 +40,12 @@
  */
 public class AudioTest {
 
+  private Audio audio;
+
+  @BeforeEach
+  void createAudioInstance() {
+    audio = new Audio();
```
IAmPramod: We should use `Audio.getInstance()` here as well.

IAmPramod:
> 1. Changed `Audio` to be a good singleton, previously because of `Audio` being bad singleton, test cases which were using static methods could have caused intermittent failures. After changes now each test case has dedicated instance of `Audio` service which provides isolation between test cases.

Why we are calling it as a `bad singleton`. It was thread-safe previously.

npathai:
> > 1. Changed `Audio` to be a good singleton, previously because of `Audio` being bad singleton, test cases which were using static methods could have caused intermittent failures. After changes now each test case has dedicated instance of `Audio` service which provides isolation between test cases.
> 
> Why we are calling it as a `bad singleton`. It was thread-safe previously.

Bad singleton because it polluted global state. It created unseen dependency between test cases and behavior would vary on how threads get interleaved. Hence the bad.

iluwatar:
Looks good to me

npathai:
@iluwatar Updated the class diagram. Ready to merge!

Issue #699, opened 2018-01-16
Intermittent failure in Balking pattern


iluwatar:
See Travis log https://api.travis-ci.org/v3/job/329176022/log.txt

iluwatar:
Ping @Rzeposlaw, care to investigate?

zhengzhanpeng:
Why this is messy when I open the Travis log?

sharmin03:
@iluwatar I would like to work on it :)

iluwatar:
Ok @sharmin03

iluwatar:
Flaky test disabled in https://github.com/iluwatar/java-design-patterns/commit/17ea0b17f68e6d20917dfd2937db28de83d9bf27. Still needs to be fixed.

PR #809, opened 2018-10-20
Fix intermittent failure in Balking pattern describe in #699

staillebois:
Fix intermittent failure

Enable the "wash" test for the Balking pattern
Solves #699

The problem here was that sometimes, the ExecutorService execute the second instruction instead of the first one.
We can see it in the log file : https://api.travis-ci.org/v3/job/329176022/log.txt
(the pool-1-thread-2 was launch before).
So the machineStateGlobal variable was set to ENABLED.

The fix proposed here is to set the machineStateGlobal variable for each execution. It will not depend anymore on the order.


iluwatar:
See @IAmPramod's comment ^

PR #814, opened 2018-10-25
Hexagonal pattern fixes

pmatejko:
Fixed wrong 'else if' (it had the same condition as previous 'if' but it should have been just 'else')
Fixed typo and changed Finnish 'saldo' word to 'balance' 
Deleted unused imports
Simplified few other things


npathai:
@pmatejko Good work :+1:

Issue #799, opened 2018-10-13
Master-Worker pattern


iluwatar:
https://docs.gigaspaces.com/sbp/master-worker-pattern.html

AnaghaSasikumar:
So, are we looking to implement this using Gigaspace or w/o it? I've never used Gigaspace before, if it's the latter, I'd like to give this a shot.

iluwatar:
No need for Gigaspace. Please go ahead @AnaghaSasikumar

AnaghaSasikumar:
Created #816 pull request

PR #816, opened 2018-10-28
Master-Worker pattern implementation #799

AnaghaSasikumar:
Master-Worker pattern implemented: pattern which helps to implement centralised parallel processing


AnaghaSasikumar:
This is including all the commits from the master branch too, so again it is very long, only the last 6 ones correspond to the master-worker pattern...what should I do to avoid that?

AnaghaSasikumar:
Please let me know what I should do, I could try to close this PR and open again if that is necessary

kezhenxu94:
I don't know whether rebasing is allowed or not here, but [git rebase](https://git-scm.com/docs/git-rebase) is a solution for this

AnaghaSasikumar:
I think that's not recommended for public repositories, might cause problems, I think the commits can be squashed and merged after review..

iluwatar:
@AnaghaSasikumar it's ok to have extra commits. They are squashed anyway when we merge. However, the pull request seems to modify other patterns too like singleton, spatial partitioning etc. Please clean it up so that there are only changes related to this pattern.

AnaghaSasikumar:
@iluwatar I deleted the change it made in singleton - there was an addition of a .checkstyle file. The rest of the changes are actually the final changes made to the spatial-partition after you posted review comments - I pulled latest changes from master before committing (i pulled from the upstream to the master then did this) - should i delete those? Will that not cause a problem? I think most of the problems are coming from the same place - that I made the changes for the first pull request directly on the master branch.
I think maybe that pull request could first be merged or rejected, according to that I'll make changes on this one too and then we can decide for this one...maybe that could work..

iluwatar:
The pull request bundles multiple issues and cannot be accepted as-is. In your fork, please create a new branch from the `master` and commit there only the changes related to one issue e.g. master-worker pattern. Then create a new pull request.

PR #821, opened 2018-11-09
Add another example to the Adapter pattern

xm-tech:
- I think it's a good idea adding one more example to the Adapter pattern
- Add another example to the Adapter pattern


iluwatar:
This pull request is about to be closed due to inactivity

PR #827, opened 2018-11-22
No Need declare App class of module pattern as a utility class

hoangnam2261:
We don't need declare App class of module pattern as a utility class.


iluwatar:
Yes, I agree @hoangnam2261. Thanks for the improvement!

PR #842, opened 2019-01-30
Fix lgtm.com alerts

joshhale:
I was inspecting this project on [lgtm.com](https://lgtm.com/projects/g/iluwatar/java-design-patterns/alerts/?mode=list), and decided to polish it a bit by fixing those nit picks.

Each alert shown on lgtm has a little (?) in the top right corner that explains it.

### DIFF
--- a/iterator/src/main/java/com/iluwatar/iterator/list/TreasureChestItemIterator.java
+++ b/iterator/src/main/java/com/iluwatar/iterator/list/TreasureChestItemIterator.java
@@ -57,9 +57,8 @@ public class TreasureChestItemIterator implements Iterator<Item> {
 
   private int findNextIdx() {
     List<Item> items = chest.getItems();
-    boolean found = false;
     int tempIdx = idx;
-    while (!found) {
+    do {
       tempIdx++;
       if (tempIdx >= items.size()) {
         tempIdx = -1;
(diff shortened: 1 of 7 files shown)


iterator/src/main/java/com/iluwatar/iterator/list/TreasureChestItemIterator.java
```diff
@@ -68,7 +67,7 @@ private int findNextIdx() {
       if (type.equals(ItemType.ANY) || items.get(tempIdx).getType().equals(type)) {
         break;
       }
-    }
+    } while(true);
```
IAmPramod: this will become an inifinite loop. you should do 
`
do {...
} while(!found)
`

joshhale: Thanks for the review!

The query on lgtm identified the loop condition here as confusing, because `found` isn't assigned in the loop. In fact the two `break` statements are the only means of exit.

joshhale: -sorry, I guess the switch from a `while` to `do` is itself a bit confusing. I guess that's just my habit to use a `do` when the condition is `(true)`.

iluwatar:
Travis build is failing due to Checkstyle errors. @joshhale please fix

iluwatar:
This pull request is about to be closed due to inactivity

PR #843, opened 2019-01-31
Fix script blocking of presentations due to address used to download remark.js

Tschis:
Sometimes the presentations would fail due to the content of remark.js being delivered through unauthenticated source.

The new source seems to work successfully.


iluwatar:
@Tschis thank you for the fix! :+1:

iluwatar:
@all-contributors please add @Tschis for content

Issue #555, opened 2017-04-01
Type Object pattern


iluwatar:
http://gameprogrammingpatterns.com/type-object.html

Imp92:
Hello @iluwatar I would like to take this pattern. Thanks!

iluwatar:
Ok @Imp92

iluwatar:
This issue is free for taking again

laubera:
Hey, @iluwatar  I would like to take this pattern. Thanks.

iluwatar:
@laubera ok

iluwatar:
@laubera  are you still implementing this?

AnaghaSasikumar:
PR #848

PR #848, opened 2019-02-09
Type object pattern #555

AnaghaSasikumar:
Type object pattern implementation

### DIFF
--- a/typeobjectpattern/src/main/java/com/iluwatar/typeobject/CandyGame.java
+++ b/typeobjectpattern/src/main/java/com/iluwatar/typeobject/CandyGame.java
@@ -1,0 +1,171 @@
+/**
+ * The MIT License
+ * Copyright (c) 2014-2016 Ilkka Seppälä
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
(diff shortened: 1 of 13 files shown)


typeobjectpattern/src/main/java/com/iluwatar/typeobject/CandyGame.java
```diff
@@ -0,0 +1,171 @@
...
+  void handleChange(int points) {
+    System.out.println("+" + points + " points!");
+    this.totalPoints += points; 
+    printGameStatus();
+  }
+  
+  void round(int timeSoFar, int totalTime) {
+    long start = System.currentTimeMillis();
+    long end = System.currentTimeMillis();
+    while (end - start + timeSoFar < totalTime && continueRound()) {
```
IAmPramod: same here. `start` and `end` are essentially the same.

AnaghaSasikumar: It's needed so that we can enter into while loop

dfparker2002:
After reading the game article that inspired this addition, I would want to see how this pattern contribution is useful. May we see an effective extension of the breed-monster idea as candy.

How would we deliver the promise of the pattern; for example, overlay a rootbeer gummy into the base candy set to dynamically grow the candy set (monster) game?

AnaghaSasikumar:
I am not sure what you mean by 'how', so do let me know if this doesn't answer your question, but I think it's a change you could probably make in the json file, assuming that we only want some kind of change in the property of the candy and not some behavioural change, i.e. maybe having this overlay would make it have more points or something like that. What I understood from the pattern was that it makes that kind of thing easier, but there will be some trade-off when it comes to behaioural change, which can't be put in the json file. It is a pretty simple pattern, I was just attracted to the idea of being able to implement so many different "types" of object using a single class and just a json file.

iluwatar:
Good job @AnaghaSasikumar :+1:

PR #864, opened 2019-03-22
Minor ReadMe Update

saritchie:
Two sentences read a little awkward, so I tweaked them.
 
1) "Design patterns are **formalized best practices that the** programmer can use" -> "Design patterns are **the best formalized practices a** programmer can use"
2) "Reusing design patterns **helps to prevent subtle issues that can** cause major" -> "Reusing design patterns **help prevent subtle issues which** cause major

Hope this helps!

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,13 +11,13 @@
 
 # Introduction
 
-Design patterns are formalized best practices that the programmer can use to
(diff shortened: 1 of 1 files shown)


iluwatar:
Thanks @saritchie :+1:

PR #869, opened 2019-04-08
Best practice when compare enum

hoangnam2261:
Dear all,
Should we apply best practice when compare enum in this case?
Reference: https://stackoverflow.com/questions/1750435/comparing-java-enum-members-or-equals
Thanks,


iluwatar:
Thanks @hoangnam2261 for this improvement!

PR #877, opened 2019-05-02
Modify UML in Observer Pattern

### LINKED ISSUE #875: Wrong UML in observer
The UML diagram presented in the Observer pattern description is wrong. 

The multiplicity between Weather and WeatherType is stated to be 0..1, but that is not the case in the code, since you can't create a Weather object without associating it with a WeatherType (the costructor sets the type to "sunny", timepasses() _changes_ that value, but there is no way to remove the type.

Therefore, the correct cardinality should be just 1.

dRadest:
Fixes #875 

- Modified ```observer.ucls``` to display the right multiplicity of 1 between Weather and WeatherType
- Replaced the newly created image with the old one in ```README.md```


iluwatar:
Thanks @dRadest :+1:

PR #890, opened 2019-06-21
fix code example containing syntax error

makaravind:
Fix code example in Ambassador pattern

- Fixed syntax error in code example in README


iluwatar:
Thanks @makaravind

Issue #898, opened 2019-07-08
A new way for Factory Method


1170300627:
I wan to attach a fork with my idea for this project

medkhelifi:
Describe your idea in this issue.

PR #906, opened 2019-07-25
Add Simple Factory Pattern(#898)

HIT1170300404:
Add Simple Factory Pattern(#898)

- Add a new java pattern, one of the factory pattern.
- It solve the #898 issue.

Pull request description

- Simple factory is also a java pattern, but it not in the project.So we add it in the project.


1170300507:
@iluwatar New Pull Request builds are failing with below error:
The command "~/bin/install-jdk.sh --target "/home/travis/oraclejdk8" --workspace "/home/travis/.cache/install-jdk" --feature "8" --license "BCL"" failed and exited with 3 during .

iluwatar:
This looks like a simplified version of Abstract Factory pattern. As such I don't think we can accept this as a new pattern.

PR #909, opened 2019-07-26
A new design-patterns : index-table

1170300133:
We created the design pattern named index-table, using an index table,which uses the town attribute in Data as the query value and the matching ID value as the stored data, and a fact table, which uses the Date class as data for storage and int as the ID value.

### DIFF
--- a/index-table/README.md
+++ b/index-table/README.md
@@ -1,0 +1,33 @@
+---
+layout: pattern
+title: Index Table
+folder: index-table
+permalink: /patterns/index-table/
+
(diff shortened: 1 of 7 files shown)


iluwatar:
Can you point to an authoritative reference where this pattern is mentioned? My first impression is that it looks like a kind of lookup table, but does not qualify as a new pattern.

iluwatar:
This pull request will be closed soon due to inactivity

iluwatar:
Closed due to inactivity

PR #910, opened 2019-07-26
Use Functional style code [abstract-document]

ibrahimAlii:
Use functional style code and direct `foreach` instead of `entrySet` one.


iluwatar:
Thanks @ibrahimAlii

Issue #903, opened 2019-07-24
CI build is failing


iluwatar:
![image](https://user-images.githubusercontent.com/582346/61822672-d0864800-ae62-11e9-9572-148ee10f8b26.png)

ptrax:
I think this is because you're using Travis CI's Xenial build environment, which seems to not allow you to install anything less than java 9 now. The choices are; upgrade to jdk 9+ or use the Trusty build environment. 

I went ahead and made a PR with the fix (using Trusty build environment). Build passes.
https://github.com/iluwatar/java-design-patterns/pull/911

PR #911, opened 2019-07-27
Change Travis CI build env. to Trusty

ptrax:
Change Travis CI Build Environment to Trusty

- This PR is to fix an issue causing the Travis CI build to fail.  The problem seems to have stemmed from the build using Travis CI's Xenial build environment, which would not allow you to install anything less than java 9.

- This is in reference to Issue #903. 

Pull request description

- The only change made here is one line of code telling Travis CI to use the Trusty build environment instead of the Xenial one.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: java
+dist: trusty   # Xenial build environment won't allow installation of Java 8
 jdk:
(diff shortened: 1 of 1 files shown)


iluwatar:
Thanks @ptrax well done :)

Issue #912, opened 2019-07-28
Web site update failing


iluwatar:
See log https://travis-ci.org/iluwatar/java-design-patterns/builds/564609653

The problem seems to be related to s3_website and Java version

ptrax:
I made a PR to take care of the Java issue, but a bunch of other errors came up with SSL errors in the HTML-Proofer and "Access key can't be null" errors in the s3_website push. I'll have to look into that.

iluwatar:
There's more information about the web site update at https://github.com/iluwatar/java-design-patterns/wiki/07.-Working-with-the-web-site

iluwatar:
@ptrax I merged the PR and it seems that the web site update system now works. See https://java-design-patterns.com it has the deployed commit information on the bottom of the screen. The access keys you mentioned are defined in the `secure` fields of [.travis.yml](https://github.com/iluwatar/java-design-patterns/blob/gh-pages/.travis.yml)

ptrax:
Thanks @iluwatar, I had just realized a few minutes ago that the access tokens weren't working since the PR came from my fork which didn't have access to those secure environment variables. Figured a post-merge build would work!

iluwatar:
Well done for fixing this, much appreciated @ptrax :+1:

PR #913, opened 2019-07-29
Change Travis CI to use Java 8 for S3_website 

ptrax:
- This PR makes Travis CI use Java 8, which requires the Trusty build environment. 
- Refer to the issue #912 

This comes from a known issue with the S3_website repo, which is not actively maintained. It seems to not support anything higher than Java 8, which makes the gh-pages build fail when Travis CI tries to use Java 11.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+dist: trusty   # Xenial build environment won't allow installation of Java 8
 env:
   global:
     - NOKOGIRI_USE_SYSTEM_LIBRARIES=true # speeds up installation of html-proofer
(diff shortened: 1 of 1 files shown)


ptrax:
Looks like the build still fails, not for the java issue now though. Bunch of SSL errors and "Access Key cant be null". Will look into it.

ptrax:
Got rid of the SSL connect errors by updating libcurl, but still have the "Access Key can not be null" on the s3_website push

iluwatar:
Thanks @ptrax for looking into this. I'll merge the changes so far.

PR #867, opened 2019-04-03
Remove the transitive dependency(junit-jupiter-api)

hoangnam2261:
Dear all,
As i see, junit-jupiter-api is is a transitive dependency of junit-jupiter-engine.
So should we remove junit-jupiter-api dependency in where we already have junit-jupiter-engine?

Reference: https://stackoverflow.com/questions/48448331/difference-between-junit-jupiter-api-and-junit-jupiter-engine

Thanks for reading,

PR #916, opened 2019-07-30
Remove the transitive dependency(junit-jupiter-api)

hoangnam2261:
- Recreate this PR for resolving the conflict for this PR:
     https://github.com/iluwatar/java-design-patterns/pull/867
- Recap:
So should we remove junit-jupiter-api dependency in where we already have junit-jupiter-engine?
Reference: https://stackoverflow.com/questions/48448331/difference-between-junit-jupiter-api-and-junit-jupiter-engine

Thank all for reading,

### DIFF
--- a/abstract-document/pom.xml
+++ b/abstract-document/pom.xml
@@ -35,11 +35,6 @@
     <dependencies>
         <dependency>
             <groupId>org.junit.jupiter</groupId>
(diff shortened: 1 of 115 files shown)


iluwatar:
@hoangnam2261 well done removing the unnecessary dependencies :+1:

hoangnam2261:
thank you, @iluwatar

PR #919, opened 2019-08-29
fix must override a superclass method

anancds:
added @override annotation when overriding a method, it helps prevent errors.


iluwatar:
Thanks @anancds

Issue #761, opened 2018-06-24
ThreadSafeDoubleCheckLocking.java: Instantiating by Reflection call will be successful if you do that firstly


DovahYol:
https://github.com/iluwatar/java-design-patterns/blob/987994f0fe5747e5524bc43c3b2000bee478c0c3/singleton/src/main/java/com/iluwatar/singleton/ThreadSafeDoubleCheckLocking.java#L43-L45
Instance will be null, and you can create a singleton object by reflection without an exception.

iluwatar:
Yes, that is true. But I think that is the way the code is supposed to work. It allows creating one and only one instance, right?

DovahYol:
Actually you can create another instance via `ThreadSafeDoubleCheckLocking.getInstance()`, or as many instances as possible via reflection.

iluwatar:
Can someone verify this?

npathai:
@iluwatar Sure. I will have a look at it.

IAmPramod:
I think we can create multiple instances of a singleton class only in case of lazy initialization as in the above example.
If we change that to eager initialization, this check will be sufficient to create multiple instances of this class.

Azureyjt:
@iluwatar May I pick this up?

iluwatar:
@Azureyjt please go ahead

Azureyjt:
PR created. Please kindly review it :) @iluwatar

PR #920, opened 2019-09-05
Fix issue #761

Azureyjt:
This PR solves issue #761 

- Add a flag to record whether the constructor is invoked for the first time. If not, throw corresponding exception.
- Add test case to test creating new instance by reflection.


iluwatar:
Thanks @Azureyjt for this fix :+1:

DeclanWang:
I think this just guarantee that the private constructor will only be invoked once but still can't prevent initiation via Reflection at the first time, right?

PR #926, opened 2019-09-19
Updating README.md file to update doc as suggested in issue#925

hbothra15:
Doc update for Adapter pattern
- Updating doc for Adapter pattern, where the doc shows different code from src code.
- This should resolve issue#925

Pull request description

- This pull request contains cosmetic change in README.md file only for ADAPTER pattern


iluwatar:
Thanks @hbothra15 for the fix :+1:

Issue #929, opened 2019-09-28
half-sync-half-async pattern need to enhance


npczwh:
In the App class of half-sync-half-async pattern, the workers in threadpool is still running when the main thread has finished. This also makes the AppTest less effective (after the test, we expect to see task results, but we get nothing because junit will terminate jvm when the main thread has finished).  
To improve on it, we can add a close method to the AsynchronousService class and call the method at the end of the main thread. I can create a PR, if needed.

PR #931, opened 2019-09-29
fixed bug #929

npczwh:
Bug fix for  (#929 )

- fixed the unclosed threads for half-sync-half-async pattern

Pull request description

- add a close method to the AsynchronousService class, to wait for all tasks to be completed and shutdown the service
- call the close method at the end of the main thread


iluwatar:
Excellent @npczwh, thanks!

Issue #922, opened 2019-09-09
Tests failing in Twin module.


snehalatapandit:
----------------------
 T E S T S
-------------------------------------------------------
Running com.iluwatar.twin.AppTest
15:50:58.934 [Thread-2] INFO com.iluwatar.twin.GameItem - draw
15:50:58.939 [Thread-2] INFO com.iluwatar.twin.BallItem - doDraw
15:50:58.939 [Thread-2] INFO com.iluwatar.twin.BallItem - move
15:50:59.189 [Thread-2] INFO com.iluwatar.twin.GameItem - draw
15:50:59.189 [Thread-2] INFO com.iluwatar.twin.BallItem - doDraw
15:50:59.189 [Thread-2] INFO com.iluwatar.twin.BallItem - move
15:50:59.439 [Thread-2] INFO com.iluwatar.twin.GameItem - draw
15:50:59.439 [Thread-2] INFO com.iluwatar.twin.BallItem - doDraw
15:50:59.439 [Thread-2] INFO com.iluwatar.twin.BallItem - move
15:50:59.680 [main] INFO com.iluwatar.twin.BallThread - Begin to suspend BallThread
15:51:00.431 [main] INFO com.iluwatar.twin.BallThread - Begin to resume BallThread
15:51:00.440 [Thread-2] INFO com.iluwatar.twin.GameItem - draw
15:51:00.440 [Thread-2] INFO com.iluwatar.twin.BallItem - doDraw
15:51:00.440 [Thread-2] INFO com.iluwatar.twin.BallItem - move
15:51:00.690 [Thread-2] INFO com.iluwatar.twin.GameItem - draw
15:51:00.690 [Thread-2] INFO com.iluwatar.twin.BallItem - doDraw
15:51:00.690 [Thread-2] INFO com.iluwatar.twin.BallItem - move
15:51:00.941 [Thread-2] INFO com.iluwatar.twin.GameItem - draw
15:51:00.941 [Thread-2] INFO com.iluwatar.twin.BallItem - doDraw
15:51:00.941 [Thread-2] INFO com.iluwatar.twin.BallItem - move
Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 2.487 sec - in com.iluwatar.twin.AppTest
Running com.iluwatar.twin.BallItemTest
15:51:01.504 [main] INFO com.iluwatar.twin.BallItem - move
[...]

PR #935, opened 2019-10-01
Fix for BallThreadTest#testResume

anirudh708:
Pull request description
I have increased the sleep time from 200 -> 300.

There is a Thread.sleep of 250 https://github.com/iluwatar/java-design-patterns/blob/master/twin/src/main/java/com/iluwatar/twin/BallThread.java#L60

So having 200 causes the test to be flaky sometimes as reported in https://github.com/iluwatar/java-design-patterns/issues/922


iluwatar:
Thanks @anirudh708 for this fix!

Issue #941, opened 2019-10-03
Java 11 support: Maven version


iluwatar:
This task is part of the Java 11 transition plan https://github.com/iluwatar/java-design-patterns/wiki/16.-Java-11-Transition

Find out what is the minimum Maven version that supports Java 11. Hopefully this information can be found on the Maven web site somewhere.

iluwatar:
![image](https://user-images.githubusercontent.com/582346/66183018-eb094880-e67f-11e9-84eb-a66b36b008b8.png)

PR #947, opened 2019-10-04
#941 Java 11 support: Maven version 3.5.0

boaglio:
_Java 11 support: Maven version 3.5.0_

There is no official note for Java 11 support, however I am positive that 3.5.0 it is the minimum version because:

- Since version 3.3 Maven requires version JDK 1.7+ ;

- Maven 3.5.0 was released in 2017, one year _before_ Java 11;

- I've tested and it works! 

_Pull request description_

- In order to test on Java 11 I needed to upgrade:
 
1.  maven-compiler-plugin up to 3.8.0 
2.  lombok up to 1.18.2 only at trampoline project due to a issue with Java 11 (got log not found with old lib)


iluwatar:
Thanks @boaglio, this information is enough to close the issue.

Issue #943, opened 2019-10-03
Java 11 support: Maven plugin upgrade


iluwatar:
This task is part of the Java 11 transition plan https://github.com/iluwatar/java-design-patterns/wiki/16.-Java-11-Transition

We need to update the Maven plugins used in the build to versions that support Java 11. The plugins include at least

- maven-compiler-plugin
- jacoco-maven-plugin
- maven-checkstyle-plugin
- maven-surefire-plugin
- maven-pmd-plugin
- license-maven-plugin

perwramdemark:
I'll fix..

iluwatar:
Thanks @perwramdemark, go ahead

PR #951, opened 2019-10-05
Upgrade of maven plugins

perwramdemark:
Fixes upgrade of maven plugins #943 

Some general code cleanup was necessary due to upgrade of PMD and checkstyle.

Also needed to add Junit 4 as a dependency due to Mockito.timout issue found here:
https://github.com/mockito/mockito/issues/152


iluwatar:
Well done @perwramdemark, thanks for this update!

Issue #549, opened 2017-03-26
"Aggregator microservices"  design pattern doesn't show exception handling 


neerajmahajan:
"Aggregator microservices"  design pattern doesn't show how exception handling would be done if the one or multiple services used are down and how client calling the aggregator microservice would cope in this situation.

while1do:
Cann't we consider 

`    try (CloseableHttpClient httpClient = HttpClients.createDefault()) {
      HttpGet httpGet = new HttpGet("http://localhost:51515/information");
      try (CloseableHttpResponse httpResponse = httpClient.execute(httpGet)) {
        response = EntityUtils.toString(httpResponse.getEntity());
      }
    } catch (IOException e) {
      LOGGER.error("Exception caught.", e);
    }`

in com.iluwatar.aggregator.microservices.ProductInformationClientImpl as example of fallback strategy ([https://blogs.oracle.com/developers/getting-started-with-microservices-part-three](url))?
Or do i not understand the task?

ddunig2:
Has this issue been resolved?

IAmPramod:
@ddunig2 I don't see a PR against this. If you want to resolve, you can go ahead.

jarpit96:
Hi, I am raising a PR for this, where ClientProtocolException is caught and  the error log statements are updated.

jarpit96:
https://github.com/iluwatar/java-design-patterns/pull/958

iluwatar:
Exception handling was improved, but the original issue still remains unsolved

iluwatar:
Under investigation by @jarpit96

jarpit96:
@iluwatar @ddunig2 To implement fallback pattern, is returning default literals/constants sufficient ?

iluwatar:
I would probably display an error message to the user in case one of the microservices does not respond or some other fault occurs

jarpit96:
#971

jarpit96:
I have added fall backs in Aggregator Service when other microservices fail. Default error messages (Product Title and Inventory) are returned to the user.

PR #971, opened 2019-10-07
Fix for Issue #549 : Add Fallbacks in Aggregator Service

jarpit96:
Changes:
Add fall backs in Aggregator Service when other microservices fail. Default error messages (Product Title and Inventory) are returned to the user.


aggregator-microservices/aggregator-service/src/main/java/com/iluwatar/aggregator/microservices/Aggregator.java
```diff
@@ -50,8 +50,21 @@
...
+    String productTitle = informationClient.getProductTitle();
+    Integer productInventory = inventoryClient.getProductInventories();
+
+    if (productTitle != null) {
+      product.setTitle(productTitle);
+    } else {
+      product.setTitle("Error: Fetching Product Title Failed"); //Fallback to error message
+    }
+
+    if (productInventory != null) {
```
iluwatar: `ProductInventoryClientImpl` does not return null but 0 when exception occurs. Perhaps returning null would be the correct way, so the change should be implemented in the client..

iluwatar:
Ping @jarpit96, some changes have been requested. Also the branch is conflicting now.

jarpit96:
Completely missed the comments, I will get on it soon. Will look at the conflicts too.

iluwatar:
Thanks @jarpit96, this looks good!

Issue #977, opened 2019-10-11
Remove maven-pmd-plugin from project


iluwatar:
We are using SonarCloud for static source code analysis so maven-pmd-plugin is included really for no particular reason. Let's remove it from the project.

Shumyk:
Hello.
Can I claim this? :)

iluwatar:
Sure @Shumyk I'll assign it to you

PR #979, opened 2019-10-11
#977 Remove maven-pmd-plugin from the project

Shumyk:
Remove maven-pmd-plugin from the project

- deleted XML parts which were responsible for adding of maven-pmd-plugin, also deleted exclude-pmd.properties.
- #977


iluwatar:
Thank you @Shumyk

Issue #970, opened 2019-10-07
Single logging framework should be enforced


iluwatar:
As aligned in https://github.com/iluwatar/java-design-patterns/wiki/10.-Technology-selections we should use Slf4j API with Logback implementation as logging framework. However, some patterns are still using log4j. In this issue we migrate them to Logback.

lauramurinova:
Hello, I would like to try and solve the issue.

iluwatar:
Thanks @lauramurinova, go ahead

iluwatar:
Ok @lauramurinova and thanks for notifying

iluwatar:
Assigned to @jjjimenez100

jjjimenez100:
PR: https://github.com/iluwatar/java-design-patterns/pull/982

PR #982, opened 2019-10-12
#970 single logging framework should be enforced

jjjimenez100:
Replaces all usages of log4j with the Slf4j API. This PR also removes references of log4j in pom and resource files.


iluwatar:
@jjjimenez100 please resolve the conflict

jjjimenez100:
Fixed by [bf2b3cb](https://github.com/iluwatar/java-design-patterns/pull/982/commits/bf2b3cbffccaf21c5493109f862ed8f7e0ea321b) Although I'm not really sure that was the best way to fix the conflict since I did it on Ghub's UI. :sweat_smile:

jjjimenez100:
> The changes look good, but log4j is still present in several places. You can try to grep the codebase like `grep -r -l "log4j"` to see where it is still referenced.

Ah right. About that, I think the only remaining are the ones under the `naked-objects` pattern. However, it seems like Apache ISIS' `withLoggingAt(..);` only accepts the concrete class of log4j's `Level`. Should I simply just remove that line?

iluwatar:
It's ok for me to remove those statements. It would be also good to scan for `System.out.println` style logging. I think I've seen some of those in the codebase.
There are some conflicts now that should be resolved.

jjjimenez100:
The only remnants of `System.out...` now would be on:

- The `Composite` pattern
- `ConsoleLoggerModule` 

since the examples has test cases asserting the output from the console (which I believe the way the example was structured to begin with)

For the log4j ones, stuffs remaining are on `persistor.properties` and all of them are already commented.

Let me know if there are any other changes I should do. :smile:

iluwatar:
Thanks @jjjimenez100 for the hard work you've done

Issue #413, opened 2016-03-31
Circuit Breaker pattern


npathai:
http://martinfowler.com/bliki/CircuitBreaker.html

The basic idea behind the circuit breaker is very simple. You wrap a protected function call in a circuit breaker object, which monitors for failures. Once the failures reach a certain threshold, the circuit breaker trips, and all further calls to the circuit breaker return with an error, without the protected call being made at all. Usually you'll also want some kind of monitor alert if the circuit breaker trips.

A production ready implementation of pattern can be found here.
https://github.com/Comcast/jrugged/blob/master/jrugged-core/src/main/java/org/fishwife/jrugged/CircuitBreaker.java

geniusgeek:
From the implementation, CircuitBreaker-State Pattern is a great pattern
compound to use for our implementation. The state pattern is to map
different state of the CircuitBreaker.  I love this pattern.

npathai:
@geniusgeek Yes we used this pattern for years in production but didnt know its name. Found it today while exploring cloud patterns. We had implemented it without official name n now i am going to refactor our production code to use it.

iluwatar:
I'll implement this pattern.

iluwatar:
http://techblog.netflix.com/2011/12/making-netflix-api-more-resilient.html

iluwatar:
https://msdn.microsoft.com/en-us/library/dn589784.aspx

iluwatar:
http://www.hascode.com/2017/02/resilient-architecture-circuit-breakers-for-java-hystrix-vert-x-javaslang-and-failsafe-examples/

MohitNarula:
Hi .. can I work on this one ?

iluwatar:
Ok @MohitNarula

iluwatar:
Issue is free

PalAditya:
Hi, wondering if I could take this issue up since it's free? Would love to contribute to this awesome repository 😄

iluwatar:
Wonderful @PalAditya, please go ahead. Looking forward to your pull request.

PR #986, opened 2019-10-12
Fix for issue #413: Circuit Breaker Pattern

PalAditya:
Fixes issue #413 by adding an implementation of circuit breaker pattern

- Added the Circuit Breaker Pattern
- Wrote test cases with nearly 100% coverage for all of the component classes (except the main class)

An example output:

output

Code Coverage Report:
Capture

Please mention whether something else needs to be done for this PR 😄


PalAditya:
> There is an error related to generating Javadocs in Travis build, although the check looks green. See the following screenshot
> 
> image

I see. I did build it locally but I use Java 8 and didn't run up against this error, and I didn't go through the Travis CI logs (😅 ). Any idea on why this could be happening?

It does seem like this is a common problem on Java 11+ builds though, as written [here](https://github.com/joel-costigliola/assertj-core/issues/1403), and I'll try to fix it as such.

Do you have any other changes in mind?

PalAditya:
Actually, I see that many modules don't seem to use the javadocs plugin, and it's not in the parent pom, so is it necessary or should I remove that from the ```pom.xml``` of ```circuit-breaker``` all together?

iluwatar:
@PalAditya yes using that plugin is not mandatory so it is ok to remove

PalAditya:
working


I believe that the previous error stemmed from this part only, and as you can see, I have fixed that, but I am getting a JaCoCo error from a completely unrelated module. Any advice on that? 😅

**Note**: Could be related to [this](https://stackoverflow.com/questions/57301580/upgrading-to-java-11-sonarqube-jacoco-with-maven-causing-errors)  or [this](https://github.com/camel-tooling/camel-language-server/issues/310)

iluwatar:
I believe the jacoco issue was fixed in https://github.com/iluwatar/java-design-patterns/issues/996. Please merge master into this branch.

circuit-breaker/pom.xml
```diff
@@ -0,0 +1,72 @@
...
+					<execution>
+						<id>report</id>
+						<phase>test</phase>
+						<goals>
+							<goal>report</goal>
+						</goals>
+					</execution>
+				</executions>
+			</plugin>
+    </plugins>
```
iluwatar: The plugins are already defined in parent pom.xml and not needed here

circuit-breaker/src/main/java/com/iluwatar/circuitbreaker/App.java
```diff
@@ -0,0 +1,86 @@
...
+   * Program entry point
+   * 
+   * @param args command line args
+   */
+  public static void main(String[] args) {
+    //Create an object of monitoring service which makes both local and remote calls
+    MonitoringService obj = new MonitoringService();
+    //Set the circuit Breaker parameters
+    CircuitBreaker circuitBreaker = new CircuitBreaker(3000, 1, 2000 * 1000 * 1000); 
+    long serverStartTime = System.nanoTime();
```
iluwatar: You can use local-variable type inference here (var)

circuit-breaker/src/main/java/com/iluwatar/circuitbreaker/CircuitBreaker.java
```diff
@@ -0,0 +1,127 @@
...
+/**
+ * The circuit breaker class with all configurations
+ */
+public class CircuitBreaker {
+  long timeout;
+  long retryTimePeriod;
+  long lastFailureTime;
+  int failureCount;
+  private final int failureThreshold;
+  private State state;
```
iluwatar: Use final where member variables don't change

circuit-breaker/src/main/java/com/iluwatar/circuitbreaker/CircuitBreaker.java
```diff
@@ -0,0 +1,127 @@
...
+  public String call(String serviceToCall, long serverStartTime) throws Exception {
+    setState();
+    if (state == State.OPEN) {
+      // return  cached response if no the circuit is in OPEN state
+      return "This is stale response from API";
+    } else {
+        // Make the API request if the circuit is not OPEN
+      if (serviceToCall.equals("delayedService")) {
+        DelayedService delayedService = new DelayedService(20);
+        String response = delayedService.response(serverStartTime);
```
iluwatar: Use local-variable type inference (var)

circuit-breaker/src/main/java/com/iluwatar/circuitbreaker/CircuitBreaker.java
```diff
@@ -0,0 +1,127 @@
...
+   */
+  CircuitBreaker(long timeout, int failureThreshold, long retryTimePeriod) {
+    // We start in a closed state hoping that everything is fine
+    this.state = State.CLOSED;
+    this.failureThreshold = failureThreshold;
+    // Timeout for the API request. Used to break the calls made to remote resource if it exceeds the limit
+    this.timeout = timeout;
+    this.retryTimePeriod = retryTimePeriod;
+    //An absurd amount of time in future which basically indicates the last failure never happened
+    this.lastFailureTime = System.nanoTime() + 1000 * 1000 * 1000 * 1000;
```
iluwatar: Could it be constant? Looks a bit cryptic

circuit-breaker/src/main/java/com/iluwatar/circuitbreaker/DelayedService.java
```diff
@@ -0,0 +1,57 @@
...
+    this.delay = 60;
+  }
+
+  /**
+   * @param serverStartTime Time at which actual server was started which makes calls to this service
+   * @return The state of the service
+   */
+  public String response(long serverStartTime) {
+    long currentTime = System.nanoTime();
+    if ((currentTime - serverStartTime) * 1.0 / (1000 * 1000 * 1000) < delay) {
```
iluwatar: The calculation here is hard to understand. We need at least comment there.

circuit-breaker/src/test/java/com/iluwatar/circuitbreaker/MonitoringServiceTest.java
```diff
@@ -0,0 +1,63 @@
...
+  }
+  
+  @Test
+  public void testRemoteResponse2() {
+    MonitoringService monitoringService = new MonitoringService();
+    CircuitBreaker circuitBreaker = new CircuitBreaker(1,1,100);
+    //Set time as current time as initially server fails
+    long serverStartTime = System.nanoTime();
+    String response = monitoringService.remoteResourceResponse(circuitBreaker, serverStartTime);
+    assertEquals(response, "Remote service not responding");
```
iluwatar: Also in tests var can be used extensively

PalAditya:
Thanks for the feedback! I implemented the changes as requested, but the Sonar Integration seems to have broken something else.

Capture

Capture

A local build with ```mvn clean install``` still passes perfectly though, so I believe the error is not with the module but somewhere else, which is causing the CI to fail. Correct me if I'm wrong 🤔 

Capture

iluwatar:
I wonder why the sonar analysis worked in my branch and master, but not here. To remedy this, I now tried to set the sonar host property in https://github.com/iluwatar/java-design-patterns/commit/47d92bbffb0e75fcbf9ea6d80dec8e15402834f7 Could you merge from the master again so we can see if it helped?

PalAditya:
Capture

After merging, I get this. It seems like it's an permission issue on your end 😅 
Maybe it is enabled for master but not PRs? 
Also, [this](https://community.sonarsource.com/t/error-youre-not-authorized-to-run-analysis-please-contact-the-project-administrator/11979) might help.

PalAditya:
Capture

Funny thing: I decided to play around with the config file and made a temporary copy of the project on my SonarQube account, and guess what, it passed (with this minimal yml on another branch)!

```yaml
language: java
dist: bionic
jdk:
- openjdk11
sudo: required

services:
- xvfb

addons:
  sonarcloud:
    organization: "paladitya"
    
script:
  # the following command line builds the project, runs the tests with coverage and then execute the SonarCloud analysis
  - mvn clean verify sonar:sonar -Dsonar.projectKey=PalAditya_java-design-patterns -Dsonar.host.url=https://sonarcloud.io

notifications:
  email:
    - adityapal.nghss@gmail.com
  webhooks:
    on_success: change  # options: [always|never|change] default: always
    on_failure: always  # options: [always|never|change] default: always
    on_start: never     # options: [always|never|change] default: always
```

I can only guess that either the encrypted key is not being read properly for PRs (I actually set the value ```SONAR_TOKEN``` in my travis settings rather than encrypting it), or maybe the ```execute-analysis``` permission needs to be set for everyone?

(Sorry if my guesses are unhelpful, but I have never worked with SonarQube before, and I find this prospect of a bug/misconfiguration quite interesting 😅 )

iluwatar:
https://docs.travis-ci.com/user/sonarcloud/#analysis-of-internal-pull-requests
The documentation indicates that it is not supposed to work across forks

PalAditya:
> https://docs.travis-ci.com/user/sonarcloud/#analysis-of-internal-pull-requests
> The documentation indicates that it is not supposed to work across forks

I see. So, what should be done next from my end? 🤔

PalAditya:
> The code in README.md explanation is not in sync with the implementation

Oh, do you mean that I should change them to use **var** too or that I shouldn't remove the comments that are in the actual code?

iluwatar:
@PalAditya can you please merge master again? It shouldn't run the sonar analysis for this PR now, if everything goes fine.

PalAditya:
@iluwatar This seems to have done the trick 😄

iluwatar:
Wonderful, thanks for your help @PalAditya :+1:

iluwatar:
Thanks for the great pattern @PalAditya, well done

Issue #988, opened 2019-10-12
Java 11 support: HttpClient


iluwatar:
This task is part of the Java 11 transition plan https://github.com/iluwatar/java-design-patterns/wiki/16.-Java-11-Transition

Patterns Aggregator Microservices and API Gateway are currently using Apache HttpClient in their implementation. Java 11 provides its own HttpClient and we should switch to that instead to get rid of unnecessary external dependencies.

Links:
https://winterbe.com/posts/2018/09/24/java-11-tutorial/#http-client

oconnelc:
I'd be interested in looking into this, can I go ahead and try?

iluwatar:
Ok @oconnelc

oconnelc:
Which version of JDK should I compile against? Should I use Oracle or Openjdk?

iluwatar:
We have just switched to openjdk11

PR #992, opened 2019-10-12
988: Replaced all of the Apache HttpClients with Java's java.net.http

oconnelc:
Replaced all of the Apache HttpClients with Java's java.net.http

This is to address ticket #988. As of Java 11 the HTTP Client provides mechanism to perform HTTP requests without the need of an external, third-party library. 

Pull request description

1. All pom.xml files that had org.apache.httpcomponents in the dependencies no longer have this as a dependency.

1. All Java files that were using any of the components of the org.apache.httpcomponents.httpClient have been modified to have equivalent code using java.net.http.


oconnelc:
I'm working on it now. I should have an update within the hour.

Issue #993, opened 2019-10-12
The pipeline project is not compiling in Eclipse


oconnelc:
After cloning the repo and opening the pipeline project Eclipse shows red errors indicating the project cannot compile:

The error is because of the layout of the project: after the src/main/java and
src/test/java. The directory was: src/main/java/com.iluwatar.pipeline. It should be src/main/java/com/iluwatar/pipeline

PR #994, opened 2019-10-12
993: Fixed the pipeines project layout so that it will load in Eclipse

oconnelc:
#993 The source directory was not working in the 2019.09 version of Eclipse. The
problem was in the layout of the project: after the src/main/java and
src/test/java, the directory was naed com.iluwatar.pipeline. It should've
been com/iluwatar/pipeline. This follows the hierarchy of all of the other
patterns.

Once these files were moved, the Pipeline project compiled without errors.


jjjimenez100:
Ah good catch. My bad on this one. (I wonder why intellij didn't convert the dot notated package name into separate dirs) Thanks!

oconnelc:
No worries. I think Eclipse is more sensitive. I thought it was weird when the project lit up.

oconnelc:
I'm working on it now. I should have an update within the hour.

Issue #984, opened 2019-10-12
Java 11 support: Start using var


iluwatar:
This task is part of the Java 11 transition plan https://github.com/iluwatar/java-design-patterns/wiki/16.-Java-11-Transition

Start using local-variable type inference (var). It is a nice example of syntactic sugar that helps reduce boilerplate code (see https://www.infoq.com/articles/upgrading-java-8-to-12/ and https://winterbe.com/posts/2018/09/24/java-11-tutorial/)

It's a huge task to do this for all the patterns at once so it is possible to create pull requests for individual patterns. This way we can complete the task gradually. The following checklist tracks which patterns have already been refactored.

[x] abstract-document
[x] abstract-factory
[x] acyclic-visitor
[x] adapter
[x] aggregator-microservices
[x] ambassador
[x] api-gateway
[x] async-method-invocation
[x] balking
[x] bridge
[x] builder
[x] business-delegate
[x] bytecode
[x] caching
[x] callback
[x] chain
[x] circuit-breaker
[x] collection-pipeline
[x] command
[x] commander
[x] composite
[x] converter
[x] cqrs
[x] dao
[x] data-bus
[x] data-locality
[x] data-mapper
[x] data-transfer-object
[x] decorator
[x] delegation
[x] dependency-injection
[x] dirty-flag
[x] double-checked-locking
[x] double-dispatch
[x] eip-aggregator
[x] eip-message-channel
[x] eip-publish-subscribe
[x] eip-splitter
[x] eip-wire-tap
[x] event-aggregator
[x] event-asynchronous
[x] event-driven-architecture
[x] event-queue
[x] event-sourcing
[x] execute-around
[x] extension-objects
[x] facade
[x] factory-kit
[x] factory-method
[x] feature-toggle
[x] fluentinterface
[x] flux
[x] flyweight
[x] front-controller
[x] guarded-suspension
[x] half-sync-half-async
[x] hexagonal
[x] intercepting-filter
[x] interpreter
[x] iterator
[x] layers
[x] lazy-loading
[x] leader-election
[x] marker
[x] master-worker-pattern
[x] mediator
[x] memento
[x] model-view-controller
[x] model-view-presenter
[x] module
[x] monad
[x] monostate
[x] multiton
[x] mute-idiom
[x] mutex
[x] naked-objects
[x] null-object
[x] object-mother
[x] object-pool
[x] observer
[x] page-object
[x] partial-response
[x] pipeline
[x] poison-pill
[...]

PR #995, opened 2019-10-12
Use local variable type inference

zbeach:
Update the following patterns with local variable type inference:
- typeobjectpattern
- unit-of-work
- value-object
- visitor

This solves part of issue #984.


iluwatar:
@zbeach please resolve the conflict and merge master

iluwatar:
Thank you for the contribution @zbeach

Issue #1001, opened 2019-10-14
Intermittent test failure in Spatial Partition pattern


iluwatar:
![image](https://user-images.githubusercontent.com/582346/66769142-09134c00-eebd-11e9-9a90-e694dac91be5.png)

afflato:
Fixed intermittent issue #1001

Smirne:
Didn't see @afflato already took this, sorry!

PR #1008, opened 2019-10-14
#1001 fix intermittent test failure - changing Rect coordinates to double.

Smirne:
#1001 QuadTreeTest was showing intermittent failures. 

The problem was due to the Rect data structure using int to store values (x, y, height, width). 

Since the QuadTree divides its space in four, we need to either to change the Rect data structure or the divide algorithm. 

I find the former more readable and clear.


iluwatar:
Thanks @Smirne for this fix!

PR #1014, opened 2019-10-17
Issue 893

hbothra15:
Static factory method
- Earlier factory was returning new Object whenever it was called to provide the child Object, now this has been replaced to provide object from static collection, this in turns help in reduction of memory foot-print, for example we ask factory method to provide Elf's Axe for Five times, earlier it used to generate 5 different object and return, now it will return Single object for each call.
- This will help in reduction of memory foot of the application

Pull request description


factory-method/src/main/java/com/iluwatar/factory/method/ElfBlacksmith.java
```diff
@@ -22,16 +22,25 @@
...
+
 /**
  * 
  * Concrete subclass for creating new objects.
  * 
  */
 public class ElfBlacksmith implements Blacksmith {
 
+  private static Map<WeaponType, ElfWeapon> ELF_ARSENAL = new HashMap<>(WeaponType.values().length);
+  {
```
iluwatar: This should be a static initializer block, so that initialization is done once only and not for every class instance

factory-method/src/main/java/com/iluwatar/factory/method/ElfBlacksmith.java
```diff
@@ -22,16 +22,25 @@
...
+    for (WeaponType type : WeaponType.values()) {
```
iluwatar: You can use Java 11 local variable type inference here (var)

factory-method/src/main/java/com/iluwatar/factory/method/ElfBlacksmith.java
```diff
@@ -22,16 +22,25 @@
...
+import java.util.Map;
+
 /**
  * 
  * Concrete subclass for creating new objects.
  * 
  */
 public class ElfBlacksmith implements Blacksmith {
 
+  private static Map<WeaponType, ElfWeapon> ELF_ARSENAL = new HashMap<>(WeaponType.values().length);
```
iluwatar: I would move the initialization of the static variable inside the static initializer block, and just declare it here

iluwatar:
@hbothra15 some changes were requested

hbothra15:
Yes @iluwatar,
Will get this done by tomorrow

factory-method/src/main/java/com/iluwatar/factory/method/ElfBlacksmith.java
hbothra15: Nope we can't as var is not applicable for method/constructor/catch formals, method return types, instance variables, or any other variable declaration

reference @ https://dzone.com/articles/java-10-lets-play-with-var

iluwatar:
Let me know when it's ready for another review

hbothra15:
It's ready for your review

factory-method/src/test/java/com/iluwatar/factory/method/FactoryMethodTest.java
```diff
@@ -95,6 +95,8 @@ public void testElfBlacksmithWithSpear() {
    * @param clazz expected class of the weapon
    */
   private void verifyWeapon(Weapon weapon, WeaponType expectedWeaponType, Class<?> clazz) {
+	System.out.println(clazz);
+	System.out.println(weapon);
```
iluwatar: Either remove these console printouts or use proper logger

hbothra15:
Changes has been made and updated the README.md file to be aligned with new code

iluwatar:
Thanks a lot @hbothra15 :+1:

PR #1018, opened 2019-10-18
Switched to embedded Google checkstyle rules

### LINKED ISSUE #1016: Checkstyle should use google_checks.xml
At the moment maven-checkstyle-plugin in main pom.xml has been configured to use custom rules defined in checkstyle.xml. However there shouldn't be many modifications compared to the builtin google_checks.xml and it would be beneficial to start using this so that own rules wouldn't have to be maintained.

kaiwinter:
Moved the configuration out of the execution to make it being used in all executions. Fixes #1016


iluwatar:
Build is failing due to exceeded maximum log length. Probably due to extensive output from Checkstyle...

kaiwinter:
Until now the travis build wasn't using the checkstyle configuration because it was in the execution I think. Now the configuration is used which means `consoleOutput` = `true`, I think that's what we are seeing here. I think it's better now as the build results are more consistent. Maybe we should set `consoleOutput` to `false`? Or go back and put the configuration in the execution? What do you think?

iluwatar:
It is outputting a lot of warnings. Those don't stop the build anyway, so I'm bit confused what is the use of them.

iluwatar:
I'm not sure about `consoleOutput=true` flag. If we set it to `false`, does it still output the reason for the failed build? If not, then I think we should keep it `true`.

kaiwinter:
Hm interesting, I moved the configuration back into the execution. Seems as the configuration still runs on travis. So it proved me wrong. The google_checks.xml contains more rules than the current one I think. So `consoleOutput` should definitely set to `false`. From the [documentation](https://maven.apache.org/plugins/maven-checkstyle-plugin/check-mojo.html) it seems `failOnViolation=true` should replace `consoleOutput` and `failsOnError`

`failsOnError`: 
> If this is true, and Checkstyle reported any violations or errors, the build fails immediately after running Checkstyle, before checking the log for logViolationsToConsole. If you want to use logViolationsToConsole, use failOnViolation instead of this.

`failOnViolation`:
> Fail the build on a violation. The goal checks for the violations after logging them (if logViolationsToConsole is true). Compare this to failsOnError which fails the build immediately before examining the output log.

kaiwinter:
Build is green ... finally

iluwatar:
This looks good. So the ruleset it is using can be found at https://github.com/checkstyle/checkstyle/blob/master/src/main/resources/google_checks.xml and it is also available in human readable format here https://google.github.io/styleguide/javaguide.html

iluwatar:
The severity of the violations in google_checks.xml is warning and that's why it's not failing the build although more than 6k issues are found.

iluwatar:
Thanks @kaiwinter for making this work :+1:

PR #1023, opened 2019-10-19
1010: Fixed the two remaining SonarCloud errors

oconnelc:
The two remaining files were still creating a Random every time the method
was called. These were missed in the previous commit because the previous
commit had fixed only one of the methods; in other words, there were
multiple methods that were creating the Random object on each call.


iluwatar:
Thanks @oconnelc for taking care of this :+1:

PR #1045, opened 2019-10-26
#1037: Fix error when building from a submodule directory

### LINKED ISSUE #1037: Error when building from submodule directories
When running `mvn clean install` from project's submodules, the following error message is outputted

```
Failed to execute goal com.mycila:license-maven-plugin:3.0:format (install-format) on project adapter: Resource license-plugin-header-style.xml not found in file system, classpath or URL: no protocol: license-plugin-header-style.xml
```

Building from the submodules should work as well as building from the multimodule root.

vanogrid:
Closes #1037 

The license-plugin-header-style.xml in the root directory was not resolved correctly when the project was built from a submodule directory, so I ended up adding a new plugin `directory-maven-plugin` which does that.

Other solutions were to use the `${maven.multiModuleProjectDirectory}` property, but I was not able to make it work, or to add a relative path property to each of the submodules `<properties><projectRoot>${parent.basedir}</projectRoot></properties>`, but this configuration would be difficult to maintain, as there are 120 submodules.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -406,6 +406,31 @@
                 </executions>
             </plugin>
 
+            <!-- Resolves ${projectRoot} to the project root directory.
+                Used by the license-maven-plugin to find the correct location of
+                license-plugin-header-style.xml when built from a submodule directory. -->
+            <plugin>
+                <groupId>org.commonjava.maven.plugins</groupId>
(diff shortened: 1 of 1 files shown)


iluwatar:
Many thanks for solving this problem for us @vanogrid. Maven really does not make it easy.

PR #1090, opened 2019-11-15
Resolves checkstyle errors for remaining m

anuragagarwal561994:
Resolves for patterns:
marker
master-worker-pattern
mediator
memento
model-view-controller
model-view-presenter
module
monad
monostate
multiton
mute-idiom
mutex


iluwatar:
Thanks @anuragagarwal561994

Issue #989, opened 2019-10-12
Package patterns into executable jars


iluwatar:
The design pattern applications are easy to run within an IDE but it would be nice to be able to run them effortlessly having only the jar files.

In this task we are going to add and configure maven-assembly-plugin to package the patterns into executable jars, so that they can later be run like `java -jar pattern.jar`.

Links:
https://www.baeldung.com/executable-jar-with-maven

eater-of-souls:
I think I'll try my hand at this

iluwatar:
Sure @eater-of-souls :smiling_imp:

iluwatar:
This issue is free to take again

hbothra15:
Before taking this up, I would Like to know do we want to add this for all patterns right
for e.g If I want to run abstract-pattern then I need to run `java -jar abstractpattern.jar`

iluwatar:
That's correct @hbothra15.

hbothra15:
Then please do assign this to me will shortly raise MR for 1 of the package and if that's good will do the same for rest

iluwatar:
Ok @hbothra15

PR #1092, opened 2019-11-16
Package patterns into executable jars setup

hbothra15:
Package patterns into executable jars

- In this PR i have tried to create the base structure in parent pom and try to implement for abstarct document, if everything looks good then can replicate the same to other packages as well.
- This PR will resolve the issue#989 partially and if approved will do the required changes for other as well

Pull request description

- This PR will have change in parent pom where we have steup the plugin managment for maven assembley plugin and leveraging the same for abstract-document pattern


hbothra15:
Fixed the conflict

hbothra15:
Hi @iluwatar ,
I would like to know, how to handle patterns which are having sub modules withing them like api-gateway, aggregator-microservices

As of now I have set plugin in each sub module to generate respective jar

iluwatar:
@hbothra15 that is the correct approach. In the microservices example we have 3 submodules that should each be runnable using `java -jar ...`. Aggregator service requires that information and inventory microservices are running so that it can fetch some information from them.

hbothra15:
If there is no conflict can we proceed with first MR, and I will be raising others in batches

iluwatar:
Thanks @hbothra15 for the first set.

iluwatar:
https://github.com/iluwatar/java-design-patterns/issues/989

Issue #1093, opened 2019-11-16
Composite Specification


iluwatar:
Specification pattern example now shows use of hard-coded and parameterized specifications. In this task we extend the example so that "composite specification" is also demonstrated. It is described in mr. Fowler's paper below.

Links:
https://martinfowler.com/apsupp/spec.pdf
https://java-design-patterns.com/patterns/specification/
https://github.com/iluwatar/java-design-patterns/tree/master/specification

MaVdbussche:
I felt like this one was coming ;) I would be glad to work on it !

iluwatar:
Wonderful @MaVdbussche, please go ahead :+1:

MaVdbussche:
@iluwatar I have created a PR for this issue at https://github.com/iluwatar/java-design-patterns/pull/1094.

PR #1094, opened 2019-11-17
Adding composite specification (Issue#1093)

MaVdbussche:
In this pull request I propose a solution to Issue#1093 (https://github.com/iluwatar/java-design-patterns/issues/1093).

I added selector classes for the three basic logic operations, and created the class AbstractSelector that allows to combine any existing selectors together  using those.
I also added a lot of explanations to the README since I feel like this approach to specification is more complex than what was present before.

I wrote unit tests in their separate class and checked the code style against Google Java Style Guide.

Please do not hesitate to send me your feedback, and I will modify my contribution as soon as I can.


iluwatar:
Good stuff @MaVdbussche. Well done! :relaxed:

Issue #1095, opened 2019-11-17
Decorator: Composition instead of Aggregation


dgruntz:
In issue #518 it was proposed to change the relationship between decorator and target (in the code example between Troll und SmartTroll) from a composition to an aggregation with the argumentation that in a composition both instances cannot live without the other one. I disagree with this argumentation and I just checked the UML spec from OMG (available at https://www.omg.org/spec/UML/2.5.1/). An example in the UML spec are wheels which are part of a car (composition), but wheels can exist without a car (in particular I just exchanged my summer wheels with winter wheels).

The spec states:

> Composite aggregation is a strong form of aggregation that requires a part object be included in at most one composite object at a time. If a composite object is deleted, all of its part instances that are objects are deleted with it.
> NOTE. A part object may (where otherwise allowed) be removed from a composite object before the composite object is deleted, and thus not be deleted as part of the composite object.

The main properties are that
1. a part may be part of at most one composite. I think that this holds for the decorator pattern, i.e. you can't decorate an instance with two different decorators, but rather the idea is that the original reference is replaced with the reference to the decorator.
2. deleting an object in one part of the graph will also result in the deletion of all objects of the subgraph below that object. But this holds for the decorator as well, i.e. if you release the smart troll instance, then the troll instance is released as well.

Therefore I recommend to switch back from aggregation to composition.

Best wishes
Dominik

iluwatar:
Thanks for the extensive explanation. Looking at the source code, I see that `ClubbedTroll` really can't exist without `Troll`. So you are right, it is a bug and should be fixed.

dgruntz:
well, don't be bedazzled by an extensive explanation ;-)

The argument raised in issue #518 was a symmetric one, i.e. it was stated that the `Troll` could also live without the `ClubbedTroll` and that therefore the association is an aggregation instead of a composition. Stated in #518, for a composition both instances, the whole and the part, cannot live without each other. Obviously, the `ClubbedTroll` cannot live without a `Troll`, but the `Troll` can obviously live without the `ClubbedTroll`. 

I argue that the _cannot live without_ argument is the wrong one, even in a composition the part could live in isolation, but the difference is that in a composition the part can be part of at most one whole, and this is what counts in this case, therefore I consider this relationship as a composition.

dgruntz:
Closed with commit #1096.

PR #1096, opened 2019-11-17
Changes the description of the decorator pattern (as described in issue #1095)

dgruntz:
As described in issue #1095, this pull request changes the description of the association between the decorator and its inner object. The current version states that this is an aggregation, but as explained in issue #1095 this should be a composition.


iluwatar:
Thanks @dgruntz, the devil is in the details :smiling_imp:

Issue #1099, opened 2019-11-17
Template Method


dgruntz:
As statet in the referenced tutorial [Template-method Pattern Tutorial](https://www.journaldev.com/1763/template-method-design-pattern-in-java)
I would add a note in the pattern description that the template method should be declared `final`. 

> To make sure that subclasses don’t override the template method, we should make it final.

And I would change the code accordingly.

iluwatar:
Yes, I agree with you.

liuc-ai:
That‘s right.

JuanfPuebla:
Asi es.

elouie-code:
@iluwatar can you please assign this to me?

elouie-code:
@iluwatar @dgruntz 
where would you want to put this new comment?
`To make sure that subclasses don’t override the template method, we should make it final.`

dgruntz:
I do not want to add this comment but rather change the code and make the template method final. This change has already been made with PR #1100, but as long as this PR is open this issue remains open as well; and the PR can't be merged yet as there is a [issue](https://github.com/mockito/mockito/issues/1826) in Mockito which has not yet been fixed.

iluwatar:
@elouie-code it's assigned to you now

ohbus:
@elouie-code is there any update regarding this issue or any progress that you made?

syfds:
hi @ohbus , @iluwatar , looks like nothing happens here, can I do a PR for this ticket?

dgruntz:
The PR #1100 is already open, but the PR can't be merged yet as there is a [issue](https://github.com/mockito/mockito/issues/1826) in Mockito which has not yet been fixed. I just checked whether the problem has been fixed with version 3.12.4 of Mockito, but this is not the case.

syfds:
I see, but in the PR #1100 there is no adjustment of README documentation (explanation why template method should be marked "final" and changing the sample code).

dgruntz:
Agreed, this could be added, and probably the class diagram needs to be adjusted as well (I don't know whether final methods are specially marked, in UML I would use a <<final>> stereotype).

syfds:
ok, so assign this ticket to me if you want to have the adjustment of the documentation:
* explanation why final is important in the Template Pattern
* adjustment of all code snippets
* adjustment of the class diagram

[...]

PR #1100, opened 2019-11-17
Declares the template methods as final (Issue 1099)

dgruntz:
As described in issue #1099 it is recommended to declare the template method as final. This pull request changes the implementation in the sample code of the following patterns:
- template method pattern
- twin pattern

The template pattern used in the callback pattern was already declared final (method `executeWith` in class `com.iluwatar.callback.Task`).

PR #1101, opened 2019-11-17
Changes the description of the template method pattern (as discussed in Issue1099)

dgruntz:
Adds a comment that the template method should be declared final (as discussed in Issue #1099).
Moreover a real-world example is added.

### DIFF
--- a/template-method/README.md
+++ b/template-method/README.md
@@ -15,11 +15,12 @@ Define the skeleton of an algorithm in an operation, deferring some
 steps to subclasses. Template method lets subclasses redefine certain steps of
 an algorithm without changing the algorithm's structure.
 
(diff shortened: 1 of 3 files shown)


dgruntz:
> I see there's documentation changes, but the actual changing the method to final is not present (anymore).

The actual change has been submitted as a separate pull request (#1100). I first only changed the code (in #1100) and then in another branch changed the documentation (#1101), but unfortunately I branched off this new branch from the wrong branch and then mannyally reverted the code change.

There is a problem with the other pull request, namely that the tests do not pass (a problem with mockito and its unability to work with final methods). Due to this problem I think it was a good idea to separate these changes in separate pull requests.

iluwatar:
Thanks @dgruntz for this contribution.

iluwatar:
#1099

PR #1102, opened 2019-11-18
Changes the description of the prototype pattern

dgruntz:
You mention in the code example how the protoype pattern can be implemented in Java with the help of Java cloning (i.e. with support of Object.clone()), but then you should also invoke Object.clone.

With the implementation

    public Sheep clone() throws CloneNotSupportedException {
      return new Sheep(name);
    }

it would not be possible (in general) to override clone() in a subclass (e.g. SpecialSheep), because invoking super.clone would return an object of the wrong type, and if we create a new instance of SpecialSheep it is not always possible to copy over the private fields.

This problem can be fixed by using the mechanism provided by Object.clone, i.e. by invoking super.clone(). This creates a shallow copy (which is just fine in this example as the type of the name field (String) is immutable.

### DIFF
--- a/prototype/README.md
+++ b/prototype/README.md
@@ -40,8 +40,12 @@ class Sheep implements Cloneable {
   public void setName(String name) { this.name = name; }
   public String getName() { return name; }
   @Override
-  public Sheep clone() throws CloneNotSupportedException {
-    return new Sheep(name);
+  public Sheep clone() {
+    try {
+      return (Sheep)super.clone();
+    } catch(CloneNotSuportedException) {
+      throw new InternalError();
+    }
   }
 }
 ```


iluwatar:
Confirmed, typically `super.clone()` should be invoked in the clone method implementation.

iluwatar:
We should change the description so that it reflects the example (there are no sheeps in the code).

dgruntz:
Beyond the example you probably should also explain the difference between Java-Cloning (a clone implementation based on super-calls up to Object.clone) and Cloning using Copy-Constructors (where each non-abstract class defines a clone method of the form

    public X clone() { return new X(this); }

and where each class (also abstract ones) provides a copy constructor which first invokes the copy constructor of the base class.

And then you should also explain the difference between deep and shallow copy.

iluwatar:
Thanks for the update @dgruntz.

PR #1103, opened 2019-11-18
Changes the implementation of the prototype pattern

dgruntz:
In the code you use the copy constructor approach to implement (in contrast to the Java-cloning approach in the description). If you follow this approach, then in every non-abstract class `X` the clone method is defined as

    public X clone() { return new X(this); }

In addition, every class (also the abstract base classes) must provide a copy constructor. I have added such constructors to the abstract base classes (which would allow to evolve these classes, i.e. to add private fields later on). These copy constructors are called form the sub classes.

The Cloneable interface does not need to be implemented, this interface is only needed when Java-Cloning is used, and the `CloneNotSupportedException` can also be dropped, as it is also only used in the context of Java Cloning (i.e. the mechanism provided by Object.clone, which is not used in this implmentation).

As a consequence of the removal of the `CloneNotSupportedException` I also had to remove the test `HeroFactoryImplTest.java` as this test checked this behavior. Actually the test was strange as it invoked clone on abstract classes which never can have instances.

I also decided to change the abstract class `Prototype` into an `Interface`. 

The contract of cloning states, that the returned object must be a new reference, but it must be equal to the origin. I added equals-implementations to all classes and added a line in the test to check whether the original and its copy are equal.


iluwatar:
@dgruntz many thanks for the improvements! :+1:

Issue #499, opened 2016-10-03
Combinator pattern


iluwatar:
https://gtrefs.github.io/code/combinator-pattern/

matthewwier:
Hi, I would like to take that, is it possible?

iluwatar:
Yes, please go ahead @matthewwier

iluwatar:
@matthewwier  are you still implementing this?

Willem-van-den-Berg:
Sure...its free

iluwatar:
@matthewwier does not seem to be active on this so @asavelev1 it is ok to start on this

iluwatar:
This issue is free to take again

besok:
I can take it if nobody minds

iluwatar:
Ok @besok

PR #1105, opened 2019-11-20
Pattern combinator

besok:
https://github.com/iluwatar/java-design-patterns/issues/499


iluwatar:
Thanks @besok for the new pattern :+1:

Issue #1047, opened 2019-10-26
Arrange the patterns into modules


iluwatar:
Currently all the patterns belong to the default unnamed module. To make them real Java modules, we need to add `module-info.java` to each of the patterns.

Links:
https://www.baeldung.com/java-9-modularity
http://tutorials.jenkov.com/java/modules.html

Progress tracking:
[x] abstract-document

saksham93:
I am working on this issue.

iluwatar:
Ok @saksham93

npczwh:
@iluwatar @saksham93  In current implementation (ex. abstract-document), we created module-info.java in the code folder (src/main/java/com.iluwatar.abstractdocument),  but my idea suggests to put it in module's source root (src/main/java). And the OpenJDK Guide (http://openjdk.java.net/projects/jigsaw/quick-start) gives the same suggestion. I'd like to know what you think of this. Should we change its path?

iluwatar:
@npczwh @saksham93 I think you're right. The module-info.java should be at the source root (src/main/java).
https://stackoverflow.com/questions/46061428/where-to-place-module-info-java-using-java-9

iluwatar:
@saksham93 are you still working on this?

saksham93:
@iluwatar  I got busy with something else but I will continue working on this as from tomorrow. Is that fine?

iluwatar:
Yes, sure @saksham93

iluwatar:
After a bit of reconsideration I've decided not to implement the module infos since it doesn't bring any benefit. In the commit above I remove the existing `module-info.java` files.

PR #1118, opened 2019-12-26
Abstract document add module info file

saksham93:
Pull request title

Abstract document add module info file. Refers to #1047 

Pull request description

Added module-info.java file in Abstract Document Design pattern.

If this is the correct way, will add it to more design patterns.


saksham93:
@iluwatar Can you please suggest if this PR is ok? Thank you.

abstract-document/src/main/java/com/iluwatar/abstractdocument/module-info.java
```diff
@@ -0,0 +1,2 @@
+module abstractdocument.modules {
+}
```
iluwatar: I would prefer format like this:
```
module com.iluwatar.abstractdocument {
}

```

iluwatar: The build is failing with `package org.slf4j is not visible
  (package org.slf4j is declared in the unnamed module, but module org.slf4j does not read it)`

To fix it you need to declare something like this inside the module-info:
`requires org.slf4j;`

saksham93: @iluwatar Thank you for your comments. I have done the required changes. Can you please suggest?

iluwatar:
https://github.com/iluwatar/java-design-patterns/issues/1047

iluwatar:
Thanks @saksham93, first one down only a few to go! :relaxed:

Issue #987, opened 2019-10-12
Java 11 support: Utilize new methods


iluwatar:
This task is part of the Java 11 transition plan https://github.com/iluwatar/java-design-patterns/wiki/16.-Java-11-Transition

This is a refactoring ticket where we start to utilize new Java 11 functionality.

1) Collecting to unmodifiable collections is possible using the new Collector for Streams operations to put the results into an immutable collection (see https://www.infoq.com/articles/upgrading-java-8-to-12/)

2) Predicate::not provides an easy way to negate predicate lambdas or method references, again reducing the boilerplate in our code (see https://www.infoq.com/articles/upgrading-java-8-to-12/)

3) New methods on Optional give even more options for coding in a functional style when using an Optional instead of having to use clumsy if statements (see https://winterbe.com/posts/2018/09/24/java-11-tutorial/)

4) Streams have new methods ofNullable, dropWhile, takeWhile (see https://winterbe.com/posts/2018/09/24/java-11-tutorial/)

5) String has new helper methods (see https://winterbe.com/posts/2018/09/24/java-11-tutorial/)

6) InputStream finally gets a super useful method to transfer data to an OutputStream (see https://winterbe.com/posts/2018/09/24/java-11-tutorial/)

7) Convenience Factory Methods for Collections make it significantly easier to create collections like lists, maps and sets (see https://winterbe.com/posts/2018/09/24/java-11-tutorial/)

It's a huge task to do this for all the patterns at once so it is possible to create pull requests for individual patterns. This way we can complete the task gradually. The following checklist tracks which patterns have already been refactored.

[x] abstract-document
[x] abstract-factory
[x] acyclic-visitor
[x] adapter
[x] aggregator-microservices
[x] ambassador
[x] api-gateway
[x] async-method-invocation
[x] balking
[x] bridge
[x] builder
[x] business-delegate
[x] bytecode
[x] caching
[x] callback
[x] chain
[x] circuit-breaker
[x] collection-pipeline
[x] command
[x] commander
[x] composite
[x] converter
[x] cqrs
[x] dao
[x] data-bus
[x] data-locality
[x] data-mapper
[x] data-transfer-object
[x] decorator
[x] delegation
[x] dependency-injection
[x] dirty-flag
[x] double-checked-locking
[x] double-dispatch
[x] eip-aggregator
[x] eip-message-channel
[x] eip-publish-subscribe
[x] eip-splitter
[x] eip-wire-tap
[x] event-aggregator
[...]

PR #1190, opened 2020-02-26
Java 11 support for #984 #987 (l)

npczwh:
Pull request title

- Java 11 support for #984 #987 (l)

Pull request description

- Updated the following patterns with local variable type inference(#984):
layers
lazy-loading
leader-election
- Checked the code in lazy-loading pattern. Nothing should be changed to Java 11 new method(#987)


iluwatar:
@npczwh many thanks for updating these patterns! :+1:

Issue #873, opened 2019-04-18
Unit of Work pattern is Too simplified


lihongjie0209:
unit-of-work/src/main/java/com/iluwatar/unitofwork/StudentDatabase.java

how can u save a complex object graph with a simple `modify` call?

should `modify` update everything in object graph?

```
public class StudentDatabase {

  public void insert(Student student) {
    //Some insert logic to DB
  }

  public void modify(Student student) {
    //Some modify logic to DB
  }

  public void delete(Student student) {
    //Some delete logic to DB
  }
}
```

cloudeswift:
It.s seems better to add a final prefix before String delete, modify, insert。
`public interface IUnitOfWork<T> {
  String INSERT = "INSERT";
  String DELETE = "DELETE";
  String MODIFY = "MODIFY";`

IAmPramod:
> It.s seems better to add a final prefix before String delete, modify, insert。
> `public interface IUnitOfWork<T> { String INSERT = "INSERT"; String DELETE = "DELETE"; String MODIFY = "MODIFY";`

@cloudeswift variables of interface are final static by default. This PR is not required https://stackoverflow.com/questions/2430756/why-are-interface-variables-static-and-final-by-default

PR #1252, opened 2020-05-31
for issue: #873

jasciiz:
It is seems better to add a final prefix before String delete, modify, insert in .\java-design-patternsnew\unit-of-work\src\main\java\com\iluwatar\unitofwork\IUnitOfWork.java .
for issue: #873


IAmPramod:
@jasciiz variables of interface are final static by default. This PR is not required https://stackoverflow.com/questions/2430756/why-are-interface-variables-static-and-final-by-default

iluwatar:
I agree with @IAmPramod - it's a fact that interface variables are static and final by default and thus the change in this PR is not needed.

iluwatar:
@all-contributors please add @IAmPramod for review

PR #1470, opened 2020-08-02
Updated ${artifactId} to ${project.artifactId}

charlesfinley:
Updated ${artifactId} to ${project.artifactId} in parent pom. Usage of ${artifactId} is deprecated.


iluwatar:
Thanks @charlesfinley 👍🏻

PR #1492, opened 2020-08-15
Fixed most reported issues by SonarCloud.

ToxicDreamz:
SonarCloud Issue fixes

- This pull request is to resolve most of the issues that SonarCloud has picked up.

Pull request description

- I've refactored most of the test cases that did not meet naming convention standards, attempted to fix the Cognitive Complexity issues within certain files, I haven't touched all of them as I was uncertain of the functionality. Mainly I've extracted parts of the original methods into their own methods to solve readability issues. I've also changed the unnecessary higher ranking access modifiers withing test cases, no changes were made to classes though, as I have no idea whether or not those methods should be used outside of the package/class itself.


ToxicDreamz:
I've fixed an issue that was causing build failures, which regarded the order of imports within a class, that's been resolved but now there seems to be a new error causing build failures, and I don't think it has something to do directly with any of the code I've modified.

I've searched for the file that it's asking for, and it is inside the project directory.

> [INFO] --- maven-install-plugin:2.4:install (default-install) @ model-view-presenter ---
> [INFO] Installing C:\Users\Toxic Dreamz\Documents\Development Projects\java-design-patterns\model-view-presenter\target\model-view-presenter.jar to C:\Users\Toxic Dreamz\.m2\repository\com\iluwatar\model-view-presenter\1.23.0-SNAPSHOT\model-view-presenter-1.23.0-SNAPSHOT.jar
> [INFO] Installing C:\Users\Toxic Dreamz\Documents\Development Projects\java-design-patterns\model-view-presenter\pom.xml to C:\Users\Toxic Dreamz\.m2\repository\com\iluwatar\model-view-presenter\1.23.0-SNAPSHOT\model-view-presenter-1.23.0-SNAPSHOT.pom
> [INFO] 
> [INFO] --- license-maven-plugin:3.0:format (install-format) @ model-view-presenter ---
> [INFO] Updating license headers...
> [INFO] ------------------------------------------------------------------------
> [INFO] BUILD FAILURE
> [INFO] ------------------------------------------------------------------------
> [INFO] Total time:  15.332 s
> [INFO] Finished at: 2020-08-15T22:15:38+04:00
> [INFO] ------------------------------------------------------------------------
> [ERROR] Failed to execute goal com.mycila:license-maven-plugin:3.0:format (install-format) on project model-view-presenter: Resource license-plugin-header-style.xml not found in file system, classpath or URL: no protocol: license-plugin-header-style.xml -> [Help 1]
> [ERROR] 
> [ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
> [ERROR] Re-run Maven using the -X switch to enable full debug logging.
> [ERROR] 
> [ERROR] For more information about the errors and possible solutions, please read the following articles:
> [ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MojoFailureException

ToxicDreamz:
Looks to me like it's a white space issue, there was a similiar whitespace issue within either the commander class or the sergeant class which caught my eye while reading to to write test cases for them, seems like I missed this one. I'll fix this right now, and push the changes.

ToxicDreamz:
Issue has been fixed and pushed, I've tested the test case locally, and it has passed.

ToxicDreamz:
In that case, I'll just keep building until I don't run into any issues. I'll request a review after those build errors are gone.

iluwatar:
You can detect Checkstyle issues faster using IDE. See instructions here: https://github.com/iluwatar/java-design-patterns/wiki/12.-IDE-instructions

ToxicDreamz:
I am using IntelliJ IDEA, and the checkstyle issues have been resolved fully, I have built the whole project locally without errors.

image

ToxicDreamz:
There is an issue that is also causing build failures but I don't exactly know of a right way to approach this. 

image

This issue is being caused by `<version></version>` tags not being present within the child modules pom.xml, I'm thinking it's causing the **child modules jupiter** dependency to default to another version instead of using the **parent modules jupiter** version, this isn't the case with all modules, some seem to be just throwing these errors and some don't. 

# **How To Recreate**
- Build the project with maven

# **Known Ways to Resolve Issue**
- Add `<version>5.5.2</version>` to the jupiter dependency within the child modules **pom.xml**.

### Example
```
<dependency>
          <groupId>org.junit.jupiter</groupId>
          <version>5.5.2</version>
          <scope>test</scope>
</dependency>
```

iluwatar (changes requested):
Partial Response build fails because you are trying to mix JUnit 4 with JUnit 5. I think the current tests in that module have been written for JUnit 4. So either upgrade everything to 5 or stick with version 4. 

You can detect the version from the imports. JUnit 4 imports look like this:
```
import static org.junit.Assert.assertEquals;
import org.junit.Before;
import org.junit.Test;
```
Whereas tests written for JUnit 5 have this kind of imports:
```
import org.junit.jupiter.api.AfterEach;
import org.junit.jupiter.api.BeforeEach;
import org.junit.jupiter.api.Disabled;
```

ToxicDreamz:
Unfortunately this is not the case, I've upgraded everything within that module to use JUnit 5, and it was still failing to build. The issue was being caused by an older version of the following dependency`

```
<groupId>org.junit.vintage</groupId>
<artifactId>junit-vintage-engine</artifactId>
```

After changing that specific dependencies version to use version **`5.4.0`**, the build was successful. Although I can't say if the issue was caused by mixing JUnit 4 and Junit 5, as I've only changed the version of the dependency after I upgraded everything to use **JUnit 5**

partial-response/pom.xml
```diff
@@ -35,18 +35,22 @@
...
             <groupId>org.junit.vintage</groupId>
             <artifactId>junit-vintage-engine</artifactId>
+            <version>5.4.0</version>
             <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>org.mockito</groupId>
-            <artifactId>mockito-core</artifactId>
+            <artifactId>mockito-junit-jupiter</artifactId>
+            <version>3.5.0</version>
```
iluwatar, line 47: Similarly here, please use dependency from parent pom.xml

pom.xml
```diff
@@ -54,6 +54,7 @@
         <annotation-api.version>1.3.2</annotation-api.version>
         <system-rules.version>1.19.0</system-rules.version>
         <urm.version>2.0.0</urm.version>
+        <mockito-junit-jupiter.version>3.5.0</mockito-junit-jupiter.version>
```
iluwatar, line 57: This should then be used in dependencyManagement section

unit-of-work/pom.xml
```diff
@@ -44,6 +44,12 @@
             <artifactId>junit-vintage-engine</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-engine</artifactId>
+            <version>5.0.0</version>
```
iluwatar, line 50: Use the one from parent pom.xml

iluwatar (changes requested):

unit-of-work/pom.xml
ToxicDreamz: If you would like me to use the version from the parent pom, then the version within the parent pom would have to be updated, as that version is causing build issues within this module.

ToxicDreamz: The other issues have been resolved and I will not push any changes just yet, as I'd like to know if you want to open a new issue regarding updating the classes that rely on the older version of **`junit-jupiter-engine`**.

PR #1493, opened 2020-08-16
updated workflows

ohbus:
## Updated Continuous Integration Workflows

- [x] Created a separate job named [Java PR Builder](https://github.com/iluwatar/java-design-patterns/compare/master...ohbus:master#diff-5ebc8c8e2c5faeb9c6834e98789d79be) for handling Pull Requests
- [x] Added caching of maven dependencies for both the jobs

### What's the improvement for having a seperate PR builder?
- PR builder job will not show the build as **failing** anymore in the [readme section badge](https://github.com/iluwatar/java-design-patterns/workflows/Java%20CI%20with%20Maven/badge.svg) for a faulty PR.

### How is caching going to help?
- Subsequent builds will be much faster reducing overall build time
- All artifacts from the previous build will be cached and need not be downloaded again for building the code.
- It will update the cache if newer dependencies are added to the **pom** files after completion of the job.

### DIFF
--- a/.github/workflows/maven-ci.yml
+++ b/.github/workflows/maven-ci.yml
@@ -1,6 +1,6 @@
 #
 # The MIT License
-# Copyright © 2014-2020 Ilkka Seppälä
+# Copyright © 2014-2019 Ilkka Seppälä
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to deal
--- a/.github/workflows/maven-ci.yml
+++ b/.github/workflows/maven-ci.yml
@@ -24,12 +24,10 @@
 # This workflow will build a Java project with Maven
 # For more information see: https://help.github.com/actions/language-and-framework-guides/building-and-testing-java-with-maven
 
-name: Java CI with Maven
+name: Java CI
 
 on:
   push:
-    branches: [ master ]
-  pull_request:
     branches: [ master ]
 
 jobs:
@@ -43,6 +41,12 @@ jobs:
       uses: actions/setup-java@v1
       with:
         java-version: 11
+    - uses: actions/cache@v1
+      with:
+        path: ~/.m2/repository
+        key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
+        restore-keys: |
+          ${{ runner.os }}-maven-
     # Some tests need screen access
     - name: Install xvfb
       run: sudo apt-get install xvfb
--- a/.github/workflows/maven-ci.yml
+++ b/.github/workflows/maven-ci.yml
@@ -1,6 +1,6 @@
 #
 # The MIT License
-# Copyright © 2014-2019 Ilkka Seppälä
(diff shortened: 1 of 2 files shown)


.github/workflows/maven-ci.yml
```diff
@@ -43,6 +41,12 @@ jobs:
       uses: actions/setup-java@v1
       with:
         java-version: 11
+    - uses: actions/cache@v1
```
iluwatar, line 44: It seems there's cache@v2 available https://github.com/actions/cache Probably we should use the latest? Also see https://docs.github.com/en/actions/configuring-and-managing-workflows/caching-dependencies-to-speed-up-workflows

.github/workflows/maven-ci.yml
```diff
@@ -1,6 +1,6 @@
 #
 # The MIT License
-# Copyright © 2014-2019 Ilkka Seppälä
+# Copyright © 2014-2020 Ilkka Seppälä
```
iluwatar, line 3: This is automatically generated using the license maven plugin, so it doesn't make much sense to update these manually for individual files.

.github/workflows/maven-pr-builder.yml
```diff
@@ -0,0 +1,64 @@
...
+    - name: Build with Maven
+      if: github.ref != 'refs/heads/master'
+      run: xvfb-run mvn clean verify
+    - name: Build with Maven and run SonarQube analysis
+      if: github.ref == 'refs/heads/master'
+      run: xvfb-run mvn clean verify org.sonarsource.scanner.maven:sonar-maven-plugin:sonar
+      env:
+        # These two env variables are needed for sonar analysis
+        GITHUB_TOKEN: ${{ secrets.REPOSITORY_ACCESS_TOKEN }}
+        SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
```
iluwatar, line 64: We need the build with SonarQube analysis in the main CI build. The other build without the analysis is for the PR builder. Probably tokens can be also removed from the PR builder.

.github/workflows/maven-ci.yml
ohbus: Okay. That's great, did not notice this for yamls in hidden folder.
Shall I let it remain as 2020, or change it back to 2019. As you said this is totally unnecessary.

.github/workflows/maven-pr-builder.yml
ohbus: Okay then I'm removing the SonarQube analysis in the PR builder

.github/workflows/maven-ci.yml
ohbus: Yes I made the changes.
Please review the changes @iluwatar 

Thanks.

iluwatar:
Many thanks @ohbus, good improvement 👍🏻

ohbus:
@iluwatar Thank you so much for guiding me.

Please do mention me if the workflow needs an update.

Thanks.

PR #1505, opened 2020-08-24
Replace System Rules with System Lambda

stefanbirkner:
System Lambda is more specific. It only wraps the part of the code that
produces the output.


iluwatar (approved):

iluwatar:
Many thanks for the update @stefanbirkner

iluwatar:
@all-contributors please add @stefanbirkner for code

stefanbirkner:
> Many thanks for the update @stefanbirkner

You're welcome. Thanks for this great resource of knowledge.

PR #1509, opened 2020-08-28
Cleaning up unnecessary code from the CI yaml and CI server version upgrade

ohbus:
- [x] Added **`-y`** flag for installing **`xvfb`**

- [x] Removed the switching using `if` that I previously forgot to eliminate from the script which was useful for using the same file for push and pull requests from **Java CI**

- [x] Removed the **`if`** checking in the Pull Requests builder as well.

- [x] upgraded CI server version from Ubuntu 18.04 to **20.04**

### DIFF
--- a/.github/workflows/maven-ci.yml
+++ b/.github/workflows/maven-ci.yml
@@ -50,7 +50,8 @@ jobs:
     # Some tests need screen access
     - name: Install xvfb
       run: sudo apt-get install -y xvfb
-    # SonarQube scan does not work for forked repositories
+    # The SonarQube analysis is only for the master branch of the main repository.
+    # SonarQube scan does not work for forked repositories try changing it to xvfb-run mvn clean verify
     # See https://jira.sonarsource.com/browse/MMF-1371
     - name: Build with Maven and run SonarQube analysis
       run: xvfb-run mvn clean verify org.sonarsource.scanner.maven:sonar-maven-plugin:sonar
--- a/.github/workflows/maven-ci.yml
+++ b/.github/workflows/maven-ci.yml
@@ -33,7 +33,7 @@ on:
 jobs:
   build:
 
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-20.04
 
     steps:
     - uses: actions/checkout@v2
--- a/.github/workflows/maven-ci.yml
+++ b/.github/workflows/maven-ci.yml
@@ -49,14 +49,10 @@ jobs:
           ${{ runner.os }}-maven-
     # Some tests need screen access
     - name: Install xvfb
-      run: sudo apt-get install xvfb
+      run: sudo apt-get install -y xvfb
     # SonarQube scan does not work for forked repositories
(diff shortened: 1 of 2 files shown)


.github/workflows/maven-ci.yml
```diff
@@ -49,14 +49,10 @@ jobs:
           ${{ runner.os }}-maven-
     # Some tests need screen access
     - name: Install xvfb
-      run: sudo apt-get install xvfb
+      run: sudo apt-get install -y xvfb
     # SonarQube scan does not work for forked repositories
     # See https://jira.sonarsource.com/browse/MMF-1371
```
iluwatar, line 54: I would extend the comment here to say that we run the SonarQube analysis only for the `master` branch.

iluwatar: And point out that there is a separate workflow for pull requests.

.github/workflows/maven-pr-builder.yml
```diff
@@ -49,9 +49,8 @@ jobs:
           ${{ runner.os }}-maven-
     # Some tests need screen access
     - name: Install xvfb
-      run: sudo apt-get install xvfb
+      run: sudo apt-get install -y xvfb
     # SonarQube scan does not work for forked repositories
     # See https://jira.sonarsource.com/browse/MMF-1371
```
iluwatar, line 54: Similarly we should here elaborate that this workflow only runs for incoming pull requests and there is a separate workflow for `master` branch.

iluwatar (changes requested):
Thanks for the improvements! I also suggest couple of additions to comments so we can make this really shine.

.github/workflows/maven-ci.yml
ohbus: Good suggestion!

.github/workflows/maven-pr-builder.yml
ohbus: Great!

ohbus:
@iluwatar Shall we also upgrade the server version from Ubuntu 18.04 to **20.04**?

iluwatar:
@ohbus why not, go ahead

ohbus:
Okay @iluwatar  Let me commit the changes and let's try that out.

ohbus:
The PR builder went smoothly! Check [here](https://github.com/iluwatar/java-design-patterns/runs/1045287215).

Let's add the the commit to master and see if it creates any problem.
If it does these will be the commit hash, to help us revert them.

- [x] [f588632](https://github.com/iluwatar/java-design-patterns/pull/1509/commits/f5886325ec8e07e994f4eeda8d7a4adb1e391f45)

- [x] [1973d1b](https://github.com/iluwatar/java-design-patterns/pull/1509/commits/1973d1bc63bc60727b5fdba0dda9070eb66f5cd9)

iluwatar (approved):

iluwatar:
Thanks @ohbus, so far looks good 👍🏻

Issue #847, opened 2019-02-08
AbstractFactory Confusing methods


markbolster12:
In the App class of the abstract-factory pattern, there are several methods which seemingly only serve to confuse readers. 
Is there a point to having a method that returns a King, from a passed in KingdomFactory?
   ```
 King getKing(final KingdomFactory factory) {
        return factory.createKing();
      }
```

iluwatar:
Apparently the method you mentioned is used in the tests. If you have a vision how to make the example clearer I would be glad to accept the pull request.

markbolster12:
One way I think that would make sense, is to alter the test. The test would instead of creating 2 factories and then testing those factories with a method never actually used (and which isn't obviously useful), the test would create the 2 factories, and then create 2 versions of App, calling createKingdom for each factory, and doing the tests like     
assertTrue(app1.getKing() instanceof ElfKing);

One other option would be to use a singular app instance and do all the tests for a particular kingdom first, then call createKingdom for the other type. Would appreciate input as to which would be more clear. I'd be happy to update it either way, and then remove the static methods from the app class.

iluwatar:
I'd vote for the second option - sounds very clear

Blackgard:
До и ш

Issue #1503, opened 2020-08-22
Introduce Lombok to get rid of boilerplate


iluwatar:
[Project Lombok](https://projectlombok.org/) enables using of annotations to generate code. Let's introduce it to the project to reduce amount of boilerplate. In the first stage we could utilize the simplest annotations like `@Getter @Setter @Slf4j @ToString @EqualsAndHashCode`. I think the more advanced ones sacrifice code readability and should be avoided.

amit1307:
Hello @iluwatar, I can work on this issue if its still unassigned?

iluwatar:
Sure @amit1307, go ahead

amit1307:
Sorry for the delay, have just started on this issue. Since it's going to be a big change so I am thinking to do it in smaller chunks, please let me know if that's okay and here is the first PR: https://github.com/iluwatar/java-design-patterns/pull/1527

ohbus:
Can I take it up @amit1307

amit1307:
Sorry @ohbus it's in progress on my side and I am kind of in the middle of it so not sure it would be a good idea for you to take it up.

carousel:
Hi everybody,

I am very familiar with Lombok and I use it often. I can work on this issue @iluwatar so feel free to assign issue to me also.

ohbus:
> Sorry @ohbus it's in progress on my side and I am kind of in the middle of it so not sure it would be a good idea for you to take it up.

Sure No problem. Go ahead!

sgarg5794:
Hi @iluwatar ,
I have 3 years of work experience in micro services development where i have worked on lombok , Can i take up this issue if it is still unassigned?

va1m:
Hi @iluwatar,
Check the PR, please.

sirNikolai:
@iluwatar I am not sure if this is honestly a good step to take. Lombok reduces boilerplate but kills legibility quite a bit, especially those unfamiliar with Lombok. One of the reasons I liked this repo is because it was very straightforward with vanilla Java, this would hinder it. Although I understand it does make writing code a little more tedious.

va1m:
> @iluwatar I am not sure if this is honestly a good step to take. Lombok reduces boilerplate but kills legibility quite a bit, especially those unfamiliar with Lombok. One of the reasons I liked this repo is because it was very straightforward with vanilla Java, this would hinder it. Although I understand it does make writing code a little more tedious.

Ah, common! 😟

[...]

PR #1517, opened 2020-09-03
847

vdlald:
Relates to [#847](https://github.com/iluwatar/java-design-patterns/issues/847)

Pull request description
* remove methods for test
* remove boilerplate code via Lombok

### DIFF
--- a/abstract-factory/pom.xml
+++ b/abstract-factory/pom.xml
@@ -31,11 +31,6 @@
 			<artifactId>junit-jupiter-engine</artifactId>
 			<scope>test</scope>
 		</dependency>
-		<dependency>
-			<groupId>org.projectlombok</groupId>
-			<artifactId>lombok</artifactId>
-			<version>1.18.12</version>
-		</dependency>
 	</dependencies>
 	<build>
 		<plugins>
--- a/abstract-factory/pom.xml
+++ b/abstract-factory/pom.xml
@@ -31,6 +31,11 @@
 			<artifactId>junit-jupiter-engine</artifactId>
 			<scope>test</scope>
 		</dependency>
+		<dependency>
+			<groupId>org.projectlombok</groupId>
+			<artifactId>lombok</artifactId>
+			<version>1.18.12</version>
+		</dependency>
 	</dependencies>
 	<build>
 		<plugins>
--- a/abstract-factory/src/main/java/com/iluwatar/abstractfactory/App.java
+++ b/abstract-factory/src/main/java/com/iluwatar/abstractfactory/App.java
@@ -77,7 +77,7 @@ public class App implements Runnable {
 
   /**
    * Creates kingdom.
-   * @param kingdomType
+   * @param kingdomType type of Kingdom
    */
   public void createKingdom(final Kingdom.FactoryMaker.KingdomType kingdomType) {
     final KingdomFactory kingdomFactory = Kingdom.FactoryMaker.makeFactory(kingdomType);
--- a/abstract-factory/src/main/java/com/iluwatar/abstractfactory/App.java
+++ b/abstract-factory/src/main/java/com/iluwatar/abstractfactory/App.java
@@ -58,10 +58,6 @@ public class App {
     setArmy(factory.createArmy());
   }
 
-  King getKing(final KingdomFactory factory) {
-    return factory.createKing();
-  }
-
   public King getKing() {
     return king;
   }
@@ -70,20 +66,12 @@ public class App {
     this.king = king;
   }
 
-  Castle getCastle(final KingdomFactory factory) {
-    return factory.createCastle();
-  }
-
   public Castle getCastle() {
     return castle;
   }
 
   private void setCastle(final Castle castle) {
     this.castle = castle;
-  }
-
-  Army getArmy(final KingdomFactory factory) {
-    return factory.createArmy();
   }
 
   public Army getArmy() {
--- a/abstract-factory/src/main/java/com/iluwatar/abstractfactory/App.java
+++ b/abstract-factory/src/main/java/com/iluwatar/abstractfactory/App.java
@@ -57,6 +57,7 @@ public class App implements Runnable {
    */
   public static void main(String[] args) {
     var app = new App();
+    app.run();
   }
 
   @Override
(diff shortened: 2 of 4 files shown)


abstract-factory/src/main/java/com/iluwatar/abstractfactory/App.java
```diff
@@ -41,12 +44,19 @@
  * and its implementations ( {@link ElfKingdomFactory}, {@link OrcKingdomFactory}). The example uses
  * both concrete implementations to create a king, a castle and an army.
  */
+@Slf4j
 public class App {
 
-  private static final Logger LOGGER = LoggerFactory.getLogger(App.class);
-
+  @Setter
```
ravening, line 50: You can use `@Data` instead of adding getter and setter to each field

vdlald: Yes, I can, but it seemed to me that the annotation `@Data` is not quite suitable for the class `App`. Usually, the classes called `App` are needed to run an application. When I look at this class, I get the feeling that there is another class called `Kingdom'. I propose to move the fields `king`, `castle`, `army` and static class `FactoryMaker` from class `App to a new class `Kingdom` (where I can use the annotation `@Data`) and in `App to use an object of class `Kingdom`. This option seems cleaner to me. What do you think about it?

ravening: Yeah you can do that as well

abstract-factory/pom.xml
```diff
@@ -31,6 +31,11 @@
 			<artifactId>junit-jupiter-engine</artifactId>
 			<scope>test</scope>
 		</dependency>
+		<dependency>
+			<groupId>org.projectlombok</groupId>
+			<artifactId>lombok</artifactId>
+			<version>1.18.12</version>
+		</dependency>
```
iluwatar, line 38: I think Lombok dependency should be introduced at the parent `pom.xml` level and then referenced here. Related to issue https://github.com/iluwatar/java-design-patterns/issues/1503 which is under construction.

abstract-factory/src/main/java/com/iluwatar/abstractfactory/App.java
```diff
@@ -41,105 +40,42 @@
...
 
-    LOGGER.info("Elf Kingdom");
-    app.createKingdom(FactoryMaker.makeFactory(KingdomType.ELF));
-    LOGGER.info(app.getArmy().getDescription());
-    LOGGER.info(app.getCastle().getDescription());
-    LOGGER.info(app.getKing().getDescription());
+  @Override
+  public void run() {
+    log.info("Elf Kingdom");
+    createKingdom(Kingdom.FactoryMaker.makeFactory(Kingdom.FactoryMaker.KingdomType.ELF));
```
iluwatar, line 61: Maybe it would be even more clear if `createKingdom` would take only the race as a parameter and handle the factory making internally?

abstract-factory/src/main/java/com/iluwatar/abstractfactory/Kingdom.java
```diff
@@ -0,0 +1,38 @@
+package com.iluwatar.abstractfactory;
+
+import lombok.Data;
+
+@Data
```
iluwatar, line 5: For the reasons described in https://github.com/iluwatar/java-design-patterns/issues/1503 I'm against using `@Data` annotation.

abstract-factory/src/test/java/com/iluwatar/abstractfactory/AbstractFactoryTest.java
```diff
@@ -23,14 +23,13 @@
 
 package com.iluwatar.abstractfactory;
 
-import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertTrue;
-
-import com.iluwatar.abstractfactory.App.FactoryMaker;
-import com.iluwatar.abstractfactory.App.FactoryMaker.KingdomType;
+import lombok.val;
```
iluwatar, line 26: For the reasons discussed in https://github.com/iluwatar/java-design-patterns/issues/1503 I'm against using Lombok `val`.

iluwatar (changes requested):

abstract-factory/src/main/java/com/iluwatar/abstractfactory/App.java
vdlald: I also think this option is more appropriate

iluwatar:
Please let me know when its ready for another review

vdlald:
> Please let me know when its ready for another review

I made the request for review via github, a request did not reach you?

vdlald:
Need to be reviewed

ravening:
> Need to be reviewed

Do you want till #1503 gets completed else the code will have boilerplate code for getters and setters

vdlald:
I removed lombok from the commit https://github.com/iluwatar/java-design-patterns/pull/1517/commits/2e36a11e24b09f43c0a6983f29191596d50c18e9.

abstract-factory/src/main/java/com/iluwatar/abstractfactory/App.java
```diff
@@ -41,84 +40,14 @@
...
-public class App {
+public class App implements Runnable {
 
-  private static final Logger LOGGER = LoggerFactory.getLogger(App.class);
+  private static Logger log = LoggerFactory.getLogger(App.class);
 
-  private King king;
-  private Castle castle;
-  private Army army;
+  private final Kingdom kingdom = new Kingdom();
```
ravening, line 47: You can refactor it to

```
Supplier<Kingdom> kingdomSupplier = Kingdom::new

...

public Kingdom getKingdom() {
    return kingdomSupplier.get();
}
```

but its upto you

vdlald: I don't quite understand why. Can you explain it in more detail?

iluwatar (approved):

iluwatar:
Many thanks for this improvement! 

@all-contributors please add @vdlald for code

vdlald:
Thanks!

PR #1530, opened 2020-09-30
Add missing generic type to Trampoline.java

carousel:
Missing generic type(T) in Trampoline.java nested object.
The test was failing due to this bug.

I have added T generic type when returning the Trampoline instance:

`return new Trampoline<T>{}`

After that test passed


iluwatar:
Thanks @carousel, the issue was fixed already in previous commits

PR #1531, opened 2020-10-01
Refactor the command pattern to use lambda functions

### LINKED ISSUE #1468: Functional implementation of Command pattern
Command pattern has been implemented in traditional style using classes. In modern Java it's possible to pass around method references. Command pattern example should be amended to demonstrate modern style implementation.

ravening:
We can leverage the lambda expressions of Java 8 onwards
to implement a command design pattern instead of traditional
non-functional way

In the traditional way, we usually have a `Command` interface and the concrete commands implement this interface
This interface will have one abstract function called `execute`
So each concrete command has to override just one function to achieve the functionality.

We know that if any interface just has one abstract method then it can be used a functional interface and can be passed around as lambda expression instead of instantiating it.

Fixes #1468


command/src/main/java/com/iluwatar/command/App.java
```diff
@@ -32,10 +32,10 @@
  * client. A command object (spell) knows about the receiver (target) and invokes a method of the
  * receiver. Values for parameters of the receiver method are stored in the command. The receiver
```
iluwatar, line 33: > Values for parameters of the receiver method are stored in the command.

I think this does not hold true anymore

command/src/main/java/com/iluwatar/command/Goblin.java
```diff
@@ -38,4 +43,20 @@ public String toString() {
...
+  }
+
+  /**
+   * changeVisibility.
+   */
+  public void changeVisibility() {
+    var visible = getVisibility() == Visibility.INVISIBLE
+            ? Visibility.VISIBLE : Visibility.INVISIBLE;
+    setVisibility(visible);
+  }
```
iluwatar, line 61: Any reason why we shouldn't put these methods in `Target` class instead?

iluwatar: Javadocs like this don't add much value

iluwatar (changes requested):
Additionally, please merge the changes from `master` branch to make the pull request build pass.

ravening:
> Additionally, please merge the changes from `master` branch to make the pull request build pass.

@iluwatar made changes. Please review again

iluwatar (approved):

iluwatar:
Thanks @ravening, this looks very good!

Issue #229, opened 2015-08-26
Page Object pattern


iluwatar:
http://martinfowler.com/bliki/PageObject.html

colinbut:
Hi @iluwatar 
I would like to contribute to your project as i have a deep interests in design patterns.

I have been using this pattern for implementing our UI tests using Selenium in the last 4-5 months so i know this pattern very well. 

Any chance i can work on this particular pattern?

iluwatar:
That is great news. Please go ahead @colinbut - your contribution is welcome!

iluwatar:
@colinbut any updates? Are you working on this?

colinbut:
@iluwatar Apologies, not had the time to start this due to important work commitments that came up lately.

Feel free to un-assign me from this issue, otherwise i will commence work on this asap.

iluwatar:
@colinbut no problem, you can continue on this when you are able.

colinbut:
Pull Request created for this pattern - #434

PR #1552, opened 2020-10-09
Added page factory design pattern

osandadeshan:
Page-Factory is a cross-platform framework that allows you to run tests on all popular browsers, because Selenium WebDriver is used to run them.

### DIFF
--- a/page-factory/pom.xml
+++ b/page-factory/pom.xml
@@ -1,0 +1,38 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.seleniumhq.selenium</groupId>
+            <artifactId>selenium-java</artifactId>
+            <version>3.14.0</version>
+        </dependency>
+        <dependency>
+            <groupId>io.github.bonigarcia</groupId>
+            <artifactId>webdrivermanager</artifactId>
+            <version>3.6.1</version>
+        </dependency>
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter</artifactId>
+            <version>RELEASE</version>
+            <scope>test</scope>
+        </dependency>
+    </dependencies>
+    <parent>
+        <groupId>com.iluwatar</groupId>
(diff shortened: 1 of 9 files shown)


iluwatar:
@osandadeshan since this change comes from outside our issue list, can you say where this testing pattern has been best described?

osandadeshan:
@iluwatar It is actually a test automation design pattern vastly used in the industry.
https://www.browserstack.com/guide/page-object-model-in-selenium

page-factory/pom.xml
```diff
@@ -0,0 +1,38 @@
...
+    <packaging>pom</packaging>
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-assembly-plugin</artifactId>
+            </plugin>
+        </plugins>
+    </build>
+</project>
```
ohbus, line 38: add an empty line at the end of the file

ohbus (changes requested):

ohbus (commented):
I cannot seem to understand one thing which is everywhere in the web _(including the reference that you have provided in the comment)_ mentions the pattern to be called as **Page Object Model** whereas you have used the term **Page Factory** design pattern.

Please clarify on the above point to help me understand you better.

Thank you.

ohbus:
> I cannot seem to understand one thing which is everywhere in the web _(including the reference that you have provided in the comment)_ mentions the pattern to be called as **Page Object Model** whereas you have used the term **Page Factory** design pattern.
> 
> Please clarify on the above point to help me understand you better.
> 
> Thank you.

@osandadeshan  What I understand from the above perspective that **Page Factory Design Pattern** is **Enhanced POM**
but on which ground.

osandadeshan:
Selenium's PageFactory, is simply an annotation that serves to delegate to Selenium the instantiation of some WebElement objects you might want to have on an object that represents a web page or a component.

As mentioned on the [Selenium documentation](https://github.com/SeleniumHQ/selenium/wiki/PageObjects):

**`There is a PageFactory in the support package that provides support for this pattern, and helps to remove some boiler-plate code from your Page Objects at the same time.`**

PageFactory serves to make things more succinct, it does not implement the Page Object design pattern out-of-the-box.

tomaszwojciechowski:
PageFactory is not a design pattern. The design pattern is Page Object, while PageFactory is a class in the Selenium framework that supports the Page Object pattern by removing the boilerplate code from the Page Object classes, instantiating those classes and setting up a lazy proxy for WebElement and List \<WebElement\> fields.

iluwatar:
I agree with @tomaszwojciechowski. We already have Page Object pattern, see https://github.com/iluwatar/java-design-patterns/issues/229.

Issue #1469, opened 2020-08-01
Functional implementation of Strategy pattern


iluwatar:
Strategy has been implemented using Java classes. In modern Java about the same can be achieved by passing methods. Let's amend the Strategy example with functional style implementation.

maheshsv:
Please assign to me thx

iluwatar:
Ok @maheshsv

maheshsv:
Do we need to implement Strategy Pattern using Lambda Expressions in Java 8

iluwatar:
Yes, that would be great addition to this pattern example

maheshsv:
ok can i create a branch and work on it, ?

iluwatar:
Go ahead @maheshsv

ignite1771:
Hi, I want to solve this issue if it is still open.
May I? @iluwatar

iluwatar:
There's a pull request from @Tubaleviao but haven't heard of him for a while

Tubaleviao:
Go ahead @ignite1771 , I'm in a quest right now

ignite1771:
Ok, I will work on this and send a PR to fix this issue.

iluwatar:
Thanks @ignite1771, looking forward to it

PR #1577, opened 2020-10-19
#1469 testing checkstyles

Tubaleviao:
Checking checkstyles to solve the issue #1469 so I changed only one line of the readme to see if the problem was already there and it has nothing to do with the changes I made in my first PR.


Tubaleviao:
Ok, the check passed so the problem was in my code.

PR #1595, opened 2020-11-07
add Page Views

xdvrx1:
This project is already big enough. But, we might want to still add the page views. This will further validate this project seeing the total views and the views per day. This can simply be added to the website also.

This just changes the `README.md` file.
 
It's just like this:

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,8 @@
 
 # Design patterns implemented in Java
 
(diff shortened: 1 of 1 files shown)


iluwatar:
Hey @xdvrx1, here I have to give the same reply as in 30-seconds-of-java. I don't see the need for yet another counter. We have the Github repository insights and the web site has its own tracking systems. Sorry, I'll reject this one.

PR #1598, opened 2020-11-19
Factory pattern : Update README.md with Real world example

ruchpeanuts:
Real world examples added for Factory pattern.


iluwatar (approved):

iluwatar:
Thanks, this is a useful addition.

@all-contributors please add @ruchpeanuts for doc

Issue #1317, opened 2020-07-07
Special Case pattern


iluwatar:
http://www.codinghelmet.com/articles/reduce-cyclomatic-complexity-special-case

ignite1771:
Hi @iluwatar I would like to work on it.

ohbus:
@ignite1771 Please go ahead!

I have assigned you the issue.

Please mention a timeline about when we can expect a Pull Request against this.

All the best!

PR #1624, opened 2021-01-10
#1317 Special Case Pattern

ignite1771:
This PR implement the e-commerce system referred in the #1317 as the example for Special Case Pattern. 

To begin with, to focus on the pattern itself, I implement the DB and maintenance lock of the e-commerce system by the singleton instance.

**Use Case**:  When a logged-in user purchases a product, different receipt view model will be showed based on different cases:
- If the transaction is successful, the default receipt will be showed.
- If the transaction is failed, the particular receipt of one of the special cases will be showed, instead of just a null object with no meaning.

To sum up, we encapsulate special cases into subclasses that provide special behaviors of the target returned interface.


special-case/etc/special-case.urm.puml
```diff
@@ -0,0 +1,119 @@
...
+ApplicationServicesImpl -->  "-domain" DomainServicesImpl
+Account --+ Db
+ApplicationServicesImpl ..|> ApplicationServices 
+DomainServicesImpl ..|> DomainServices 
+DownForMaintenance ..|> ReceiptViewModel 
+InsufficientFunds ..|> ReceiptViewModel 
+InvalidUser ..|> ReceiptViewModel 
+OutOfStock ..|> ReceiptViewModel 
+ReceiptDto ..|> ReceiptViewModel 
+@enduml
```
ohbus, line 119: Add a new line

special-case/pom.xml
```diff
@@ -0,0 +1,22 @@
...
+    <dependencies>
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-engine</artifactId>
+            <scope>test</scope>
+        </dependency>
+    </dependencies>
+
+
+</project>
```
ohbus, line 22: Add a new line

special-case/src/test/java/com/iluwatar/specialcase/AppTest.java
```diff
@@ -0,0 +1,16 @@
...
+/**
+ * Application test.
+ */
+public class AppTest {
+
+  @Test
+  void shouldExecuteWithoutException() {
+    assertDoesNotThrow(() -> App.main(new String[]{}));
+  }
+}
```
ohbus, line 16: Add a new line

ohbus (changes requested):

special-case/src/test/java/com/iluwatar/specialcase/AppTest.java
ignite1771: Thanks. Solve it in 43a9fab.

special-case/pom.xml
ignite1771: Thanks. Solve it in 43a9fab.

special-case/etc/special-case.urm.puml
ignite1771: Thanks. Solve it in 43a9fab.

ohbus (approved):

ohbus:
@ignite1771 Thank you for contributing to this repository.

It was a great pull request overall.

ohbus:
@all-contributors please add @ignite1771 for code

PR #1574, opened 2020-10-17
Updated Java 8 lambda functions

Tubaleviao:
Fixes #1469

I added parameters and return values to the classes to show a little more about the advantages of using the Strategy Pattern. Plus I added the lambda functions to the README file.

I tested using Eclipse.

PR #1631, opened 2021-01-18
#1469 Add lambda expressions implementation for Strategy Pattern

ignite1771:
This PR implements the functional implementation of Strategy Pattern referred in the #1469.

**Use Case**:  DragonSlayer can change its strategy by passing methods, which can be achieved by lambda expressions (after Java 8) combined with a enum.

To sum up, we don't use java classes to implement Strategy but passing methods.

Thanks to the discussion in #1574 for providing some ideas.


ohbus (approved):

ohbus:
Thanks @ignite1771 for the improvement.

PR #1632, opened 2021-01-19
update gitignore for checkstyle

ohbus:
Tried with Eclipse Java EE 2020-09 on Linux and 2020-12 on Windows platform.

**.checkstyle** files are being tracked which should not be


iluwatar:
Makes sense, thanks @ohbus

Issue #1655, opened 2021-02-14
build: Introduce Maven Wrapper


ohbus:
For moving towards the goal of not having Maven installed in the system, we are introducing the maven wrapper.

This is going to help contributors from installing maven in their system and building code using the command line as well as not worry about which maven version to use for building this project. _(Though we never got such reports of the build misbehaving due to the maven version)_

We will need to setup the CI build as well as the Wiki for the issue to be closed successfully.

</br>

✔️  Checks before closing this issue.
- [x] Introduce Maven Wrapper
- [x] Update CI to use the Wrapper instead of the default Maven provided by Actions.
~~Update Wiki to use Wrapper to build projects.~~

PR #1656, opened 2021-02-14
#1655: added maven wrapper

ohbus:
### Added Maven Wrapper

These files are from Maven wrapper. It works similarly to the Gradle wrapper.

This allows you to run the Maven project without having Maven installed and present on the path. It downloads the correct Maven version if it's not found (as far as I know by default in your user home directory).

The `mvnw` file is for Linux (bash) and the `mvnw.cmd` is for the Windows environment.

Partly addresses #1655 

- [x] Maven Binary Version `3.6.3`
- [x] Maven Wrapper Version `0.5.6`


iluwatar (approved):

iluwatar:
Wonderful, I think using the wrapper is a good idea. Thanks @ohbus 👍🏻

Issue #1625, opened 2021-01-12
Turkish Translation


demirhalil:
Hello I'd like to contribute this project by adding Turkish translation. Do you agree? @iluwatar @xiaod-dev

@all-contributors can you please add me for translation ?

ohbus:
Hey @demirhalil , we currently do not have turkish translation at the moment.

Would you like to work on it?

demirhalil:
Hey @ohbus yeah I'd like to work on it.

ohbus:
@demirhalil We are happy to have you on board.

Looking forward to your first Pull Request.

Make sure to go through our [Wiki](https://github.com/iluwatar/java-design-patterns/wiki) section for understanding this better. If you still need any help, please reach out to us.

PR #1665, opened 2021-03-01
#1625 add Turkish translation for singleton design pattern

demirhalil:
#1625
Pull request description

Turkish translation is added for singleton design pattern


ohbus (approved):
Thank you @demirhalil for this translation.

PR #1670, opened 2021-03-06
updated Random() to SecureRandom()

ohbus:
As reported by SonarCloud use should use `SecureRandom()` instead of `Random()`


ohbus:
If there are any static code errors, I will fix in the meantime!

Thanks.

iluwatar (approved):

iluwatar:
Good improvement, thanks @ohbus 👍🏻

PR #1676, opened 2021-03-14
refactor: added license headers to active object pattern

ohbus:
Added license headers to active object pattern


iluwatar (approved):

iluwatar:
Thanks @ohbus, for good refactoring!

PR #1677, opened 2021-03-16
Exclude junit4 from deps and fix tests for naked-objects module

PakhomovAlexander:


ohbus (changes requested):
LGTM 🎉 

Just a request for you to remove the static code errors

Issue #590, opened 2017-06-11
Explanation for design patterns


iluwatar:
We want to present the design patterns better as the current does not reach its goal of explaining how the pattern works in a simplest possible way. I'm hugely inspired by https://github.com/kamranahmedse/design-patterns-for-humans repository.

Let's do this by adding 'Explanation' section to the patterns' `README.md` file. For an example see http://java-design-patterns.com/patterns/bridge/.

Let's track the issue progress here with the following checklist that indicates the patterns where explanation section is already available.

- [x] abstract-document
- [x] abstract-factory
- [x] active-object
- [x] acyclic-visitor
- [x] adapter
- [x] aggregator-microservices
- [x] ambassador
- [x] api-gateway
- [x] arrange-act-assert
- [x] async-method-invocation
- [x] balking
- [x] bridge
- [x] builder
- [x] business-delegate
- [x] bytecode
- [x] caching
- [x] callback
- [x] chain
- [x] circuit-breaker
- [x] cloud-static-content-hosting
- [ ] collection-pipeline
- [ ] combinator
- [x] command
- [ ] commander
- [x] composite
- [x] composite-entity
- [x] converter
- [ ] cqrs
- [x] dao
- [ ] data-bus
- [ ] data-locality
- [ ] data-mapper
- [x] data-transfer-object
- [x] decorator
- [ ] delegation
- [x] dependency-injection
- [ ] dirty-flag
- [ ] double-buffer
- [ ] double-checked-locking
- [ ] double-dispatch
- [ ] eip-aggregator
- [ ] eip-message-channel
- [ ] eip-publish-subscribe
- [ ] eip-splitter
- [ ] eip-wire-tap
- [ ] event-aggregator
- [ ] event-based-asynchronous
- [ ] event-driven-architecture
- [ ] event-queue
- [ ] event-sourcing
- [x] execute-around
- [ ] extension-objects
- [x] facade
- [x] factory
- [ ] factory-kit
- [x] factory-method
- [ ] fan-out-fan-in
- [ ] feature-toggle
- [x] filterer
- [x] fluentinterface
- [ ] flux
- [x] flyweight
- [ ] front-controller
- [x] game-loop
- [ ] guarded-suspension
- [ ] half-sync-half-async
- [ ] hexagonal
- [ ] intercepting-filter
- [x] interpreter
- [x] iterator
- [x] layers
- [ ] lazy-loading
- [ ] leader-election
- [ ] leader-followers
[...]

PR #1686, opened 2021-03-20
Explanation for business delegate

iluwatar:
As part of issue #590, this pull request

- Refactors the business delegate example to make it a bit more interesting
- Adds explanation and other sections to `README.md`
- Updates the URM generated class diagram and removes the old hand-made ones


ohbus (approved):
Good improvement! Thank you so much @iluwatar 🎉

PR #1689, opened 2021-03-22
Fix MVVM pattern readme

iluwatar:
Noticed that after merging MVVM it has some display problems on the website https://java-design-patterns.com/patterns/model-view-viewmodel/. In this PR:

- Fixes minor problems in MVVM pattern `README.md` to display it correctly on the website


ohbus (approved):
Thanks for the improvement @iluwatar

PR #1698, opened 2021-04-08
Improvement: multiple-assertions unit test into parameterized one

eas5:
Problem:
A test method with many individual assertions stops being executed on the first failed assertion, which prevents the remaining ones' execution. In the refactored method, the difference between the assertions lies within different arguments only.

Solution:
Parameterized tests make it possible to run a test multiple times, with different arguments, as individual and independent tests. This way, we were able to make 1 original test with 9 assertions become 9 independent ones, one for each enumeration element. 

Additional change:
Maven dependency concerning the JUnit params library was also added.

### DIFF
--- a/multiton/pom.xml
+++ b/multiton/pom.xml
@@ -38,6 +38,11 @@
       <artifactId>junit-jupiter-engine</artifactId>
       <scope>test</scope>
     </dependency>
+      <dependency>
+          <groupId>org.junit.jupiter</groupId>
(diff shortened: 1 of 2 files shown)


ohbus (approved):
LGTM

Let us wait to fix the build problem before we could merge this into master.

> We need #1699 to be rebased into this PR after it is has been merged

ohbus:
Thank you so much @eas5 for the improvement 🎉 

@all-contributors please add @eas5 for code

PR #1708, opened 2021-04-20
Translated the strategy pattern into Korean

keunyop:
- Spelling correction and human-readable paraphrase on ko/README.md. Previously, it was translated using a machine translator.
- Translated the strategy pattern into Korean.


ohbus (changes requested):
Thank you so much for the improvement @keunyop 

We would only encourage you to submit the translations of the README files and not the actual source code.

keunyop:
> Thank you so much for the improvement @keunyop
> 
> We would only encourage you to submit the translations of the README files and not the actual source code.

I'll do. Thank you.

ohbus (approved):

Issue #1260, opened 2020-07-07
Compose Method pattern


iluwatar:
https://www.informit.com/articles/article.aspx?p=1398607

wangihzyue:
Hi, @iluwatar , I also want to try this pattern. Is it ok? I suppose that this is relatively easy, so I want to try it.

ohbus:
Sure thing @wangihzyue

vishhhhhhhh:
I am happy to have a go at this task, please assign it to me :)

iluwatar:
Yes @vishhhhhhhh, it's assigned to you now

PR #1713, opened 2021-04-22
#1260 add Compose Method pattern

wangihzyue:
- Resolves issue #1260 
- The idea of the pattern is to transform the logic into a small number of intention-revealing steps at the same level of detail.


ohbus (changes requested):
Thanks for the contribution, please add this module to the parent pom to start building the code.

ohbus (changes requested):
Looks good but please **increase test coverage** as I can see there are a lotion untested methods.

And additionally, remove all code smells that it has.

_PS. Is there any particular reason why test coverage information is not working for this PR?_

compose-method/README.md
```diff
@@ -0,0 +1,139 @@
+---
+layout: pattern
+title: Compose-method
+folder: compose-method
+permalink: /patterns/compose-method/
+categories: Structural
+tags:
+ - Gang of Four
```
iluwatar, line 8: It's not in GoF book. I suggest using `Decoupling` tag. https://github.com/iluwatar/java-design-patterns/wiki/07.-Categories-and-Tags

compose-method/README.md
```diff
@@ -0,0 +1,139 @@
...
+
+> Composed Method is a small, simple method that you can understand in seconds.
+
+Wikipedia says
+
+> In software engineering, composed method is to divide your program into methods that perform one identifiable task. Keep all of the operations in a method at the same level of abstraction. This will naturally result in programs with many small methods, each a few lines long.
+
+**Programmatic Example**
+
+Refactoring to Patterns: Simplification | Compose Method | InformIT
```
iluwatar, line 31: Is it referring to the source of the example? Maybe add a link here to make it really easy for the user.

iluwatar: Maybe we could also explain that in the example we refactor a complex method into the composed method pattern

compose-method/pom.xml
```diff
@@ -0,0 +1,35 @@
...
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-assembly-plugin</artifactId>
+            </plugin>
+        </plugins>
+    </build>
+    <properties>
+        <maven.compiler.source>11</maven.compiler.source>
+        <maven.compiler.target>11</maven.compiler.target>
+    </properties>
```
iluwatar, line 33: These lines are probably not needed. They should come from the parent pom.xml

compose-method/src/main/java/com/iluwatar/composemethod/App.java
```diff
@@ -0,0 +1,42 @@
+package com.iluwatar.composemethod;
+
+import java.util.Random;
+import lombok.extern.slf4j.Slf4j;
+
+/**
+ * add original method is rough, and the new add method is using the composed method.
+ * {@link Node} is the node for list, and this term is to add the positive integer in
+ * each node's integer list
```
iluwatar, line 9: Maybe first add a short explanation of the pattern and then explain how this example implements it

compose-method/src/main/java/com/iluwatar/composemethod/App.java
```diff
@@ -0,0 +1,42 @@
...
+  /**
+   * Program entry point.
+   *
+   * @param args command line args.
+   */
+  public static void main(String[] args) {
+    //initialize a list.
+    Node head = new Node();
+    LOGGER.info("head: {}", head);
+    Random random = new Random();
```
iluwatar, line 22: You can use `var` keyword here. Please check elsewhere too.

iluwatar (changes requested):

compose-method/README.md
iluwatar: Since this pattern is a small one, I would use `Idiom` in the categories field

compose-method/src/main/java/com/iluwatar/composemethod/App.java
ohbus: Also, use SecureRandom

wangihzyue:
transfer to new pr [https://github.com/iluwatar/java-design-patterns/pull/1759](https://github.com/iluwatar/java-design-patterns/pull/1759)

PR #1727, opened 2021-04-26
Translated the balking pattern into Chinese

Al-assad:


ohbus (approved):
Thanks, @Al-assad for your contribution.

PR #1737, opened 2021-04-30
Updated the docker image

### LINKED ISSUE #1736: Update next-generation circleci convenience images
The next-generation convenience images in this section were built from the ground up with CI, efficiency, and determinism in mind. Here are some of the highlights:

**Faster spin-up time** - In Docker terminology, these next-gen images will generally have fewer and smaller layers. Using these new images will lead to faster image downloads when a build starts, and a higher likelihood that the image is already cached on the host.

**Improved reliability and stability** - The current images are rebuilt practically every day with potential changes from upstream that we can’t always test fast enough. This leads to frequent breaking changes, which is not the best environment for stable, deterministic builds. Next-gen images will only be rebuilt for security and critical bugs, leading to more stable and deterministic images.

> https://circleci.com/docs/2.0/circleci-images/#next-generation-convenience-images
> https://hub.docker.com/u/cimg
> https://github.com/CircleCI-Public/cimg-overview
> https://hub.docker.com/r/cimg/openjdk
> https://discuss.circleci.com/tag/convenience-images

sapsinh:
Resolves #1736

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -26,7 +26,7 @@ version: 2
 jobs:
   sonar-pr:
     docker:
-      - image: cimg/openjdk:11.0.6-node
+      - image: cimg/openjdk:11.0-node
     steps:
     - checkout
     - restore_cache:


iluwatar:
Hey what's going on here, is anybody working on this or should we close the PR?

Issue #1303, opened 2020-07-07
Mapper pattern


iluwatar:
https://books.google.fi/books?id=vqTfNFDzzdIC&pg=PA473#v=onepage&q&f=false

bravogalano:
@iluwatar Hi, can I take this issue?

ohbus:
⭐ Sure thing! Thanks for your interest in our project 😃 

✔️ Please mention a 📆 timeline 🕙 for when can we expect 🤔 a Pull Request against this issue.

bravogalano:
I suppose tomorrow I can make a pull request.

shibicr93:
Hello @iluwatar , I'd like to work on this issue. If assigned, I can open a PR in 2 days.

iluwatar:
Assigned to you @shibicr93

shibicr93:
to be fixed in https://github.com/iluwatar/java-design-patterns/pull/2064.

PR #1772, opened 2021-05-22
#1303 add mapper pattern

bravogalano:
Resolve issue#1303
Mapper pattern is to set up communications between two subsystems that still need to stay ignorant of each other.


iluwatar (changes requested):
Thank you for the pull request! The main issue at the moment is that the example is very hard to understand. It might be because of unclear naming of the classes or insufficient explanation. There are also problems with data structures and interfaces further complicating the implementation.

bravogalano:
OK, I will change it later.

Issue #1012, opened 2019-10-15
SonarCloud reports issues


iluwatar:
The project is using SonarCloud static code analysis. The latest results are showing that it has found several blockers and critical severity code smells (major, minor, and info level we ignore). In this task, those are fixed or marked as false positives. To mark false positives, SuppressWarnings annotation should be used in code as explained in the linked documentation.

Links:
[Blocker and critical severity code smells in SonarCloud](https://sonarcloud.io/project/issues?id=iluwatar_java-design-patterns&resolved=false&severities=BLOCKER%2CCRITICAL&types=CODE_SMELL)
[Sonar false positives](https://docs.sonarqube.org/latest/analysis/languages/java/#JavaFAQ-HowtoremoveFalse-Positiveissues)

***

Note: When Submitting a PR please provide a **hyperlinked** list of all the issues that you are issuing a fix for. [_See this example_](https://github.com/iluwatar/java-design-patterns/pull/1833#issue-1015516036)

ykayacan:
Can I take this issue?

iluwatar:
Sure @ykayacan

iluwatar:
This issue is free for taking again.

ToxicDreamz:
I would like to take this issue.

ToxicDreamz:
Through out checking the files causing issues on SonarCloud, I've encountered test cases that are not properly named to the functionality that they are to be tested against. 

Example`

In-correct naming convention of a test method.
> `  @Test
>   void test() throws Exception {
>     App.main(new String[]{});
>   }`

Correct naming convention of a test method.

> `@Test
>   public void shouldExecuteAppWithoutException() {
>     assertDoesNotThrow(() -> App.main(null));
>   }`

A test method should describe what its testing directly in the naming, I will rename the test cases that I find throughout the refactoring process to match proper convention rules, if the test case is understandable as to what it's testing, but I would encourage that these test cases be looked at and properly named for other contributors making changes to certain files that the test cases test against, otherwise it causes a lot of confusion as to whether said contributor is testing the right thing.

iluwatar:
Ok @ToxicDreamz, assigned to you. I agree with your comment regarding test naming.

ToxicDreamz:
I've already finished with the issue. 

Request to run another SonarCloud report after merging, to see what else remains, and whether or not they can be modified so issues do not come up again.

anuragagarwal561994:
@iluwatar happy to be back again :)
[...]

PR #1784, opened 2021-06-07
#1012 - Resolve Sonar report: missing assertions in several AppTest c…

samuelpsouza:
Resolve part of issue #1012
Some blocking issues found in the Sonar report are related to the lack of assertions in AppTest classes. The changes provide those assertions. Also, a few methods were renamed to a more precise name.

### DIFF
--- a/parameter-object/src/test/java/com/iluwatar/parameter/object/AppTest.java
+++ b/parameter-object/src/test/java/com/iluwatar/parameter/object/AppTest.java
@@ -23,19 +23,16 @@
 
 package com.iluwatar.parameter.object;
 
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
+
 import org.junit.jupiter.api.Test;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 /**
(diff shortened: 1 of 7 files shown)


parameter-object/src/test/java/com/iluwatar/parameter/object/AppTest.java
```diff
@@ -23,19 +23,16 @@
...
 /**
  * Application test
  */
 class AppTest {
-  private static final Logger LOGGER = LoggerFactory.getLogger(AppTest.class);
-
   @Test
   void shouldExecuteApplicationWithoutException() {
-    App.main(new String[]{});
-    LOGGER.info("Executed successfully without exception.");
```
ohbus, line 39: Any reasons for removing logging? Was it showing any issues on sonar?

samuelpsouza: The purpose of the logging was to assert that the code executed without errors. Since there is an aproprieted assertion for that, there is no need for this logging.

ohbus (approved):

PR #1786, opened 2021-06-14
Simplify switch statement using Enums

ravening:
If we add more enums then we need to add more
switch cases which is not a good coding convention.
Instead using Enums abstract function feature to
provide the definition

Add a new abstract function in Enum to create a new Potion object rather
than having several switch cases to create one.

### DIFF
--- a/flyweight/README.md
+++ b/flyweight/README.md
@@ -18,20 +18,20 @@ Use sharing to support large numbers of fine-grained objects efficiently.
 
 Real world example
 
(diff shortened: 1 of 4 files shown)


iluwatar (changes requested):
Nice trick with the enum and I agree it makes the implementation a notch nicer. Please reflect the changes you've made in the flyweight `README.md` as well.

iluwatar:
Closed due to inactivity

Issue #1294, opened 2020-07-07
Domain Model pattern


iluwatar:
https://inviqa.com/blog/architecture-patterns-domain-model-and-friends

AndriyPyzh:
@iluwatar  I would like to implement that.

iluwatar:
Ok @AndriyPyzh

AndriyPyzh:
@iluwatar I raised a PR, please review.

PR #1795, opened 2021-06-28
Added Domain Model pattern

AndriyPyzh:
Domain Model pattern

- Resolves issue #1294 
- A Domain Model is an object model of the domain that incorporates both behavior and data.

Description

- Implement domain model pattern 
- corrected main class name in pom.xml of strategy pattern
- minor update of "Related patterns" in readme file of transaction script pattern


tedyoung:
Domain Model **is not** a Design Pattern, it's a way of modeling and structuring code in complex systems (I'd even say that it's just the right way to do object-oriented programming). Even if it were a pattern, this code directly contradicts the intent by having the "domain objects" directly reference the persistence objects (DAO).

iluwatar:
> Domain Model **is not** a Design Pattern, it's a way of modeling and structuring code in complex systems (I'd even say that it's just the right way to do object-oriented programming). Even if it were a pattern, this code directly contradicts the intent by having the "domain objects" directly reference the persistence objects (DAO).

@tedyoung thanks for the comment! The domain model is listed as one of the domain logic patterns in Fowler's [Patterns of Enterprise Application Architecture](https://www.amazon.com/Patterns-Enterprise-Application-Architecture-Martin/dp/0321127420/ref=sr_1_1?crid=1Y3ZS7XOAUJNU&dchild=1&keywords=patterns+of+enterprise+application+architecture&qid=1625910501&s=books&sprefix=patterns+of+enter%2Caps%2C266&sr=1-1). I definitely think that it's a valid design pattern, but I also understand the confusion around the terms. For example, the book by Evans, [Domain-Driven Design](https://www.amazon.com/Domain-Driven-Design-Tackling-Complexity-Software-dp-0321125215/dp/0321125215/ref=mt_other?_encoding=UTF8&me=&qid=1625910547) is all about designing domain models for applications. The concept is so important and goes so deep into the core of software development that it may feel like something larger than a design pattern.

domain-model/src/test/java/com/iluwatar/domainmodel/CustomerDaoImplTest.java
```diff
@@ -0,0 +1,161 @@
...
+            .build();
+  }
+
+  @AfterEach
+  void tearDown() throws SQLException {
+    TestUtils.deleteSchema(dataSource);
+  }
+
+  @Test
+  void findByNameTest() throws SQLException {
```
iluwatar, line 79: We don't need to have the "test" suffix in the test method names. The test method name should always describe what is being tested.

iluwatar: See some examples here https://stackoverflow.com/questions/155436/unit-test-naming-best-practices

iluwatar (changes requested):

AndriyPyzh:
@iluwatar thanks for the review! I updated PR according to your suggestions.

domain-model/README.md
```diff
@@ -0,0 +1,314 @@
+---
+layout: pattern
+title: Domain Model
+folder: domain-model
+permalink: /patterns/domain-model/
+categories: Structural
+tags:
+ - Data access
```
iluwatar, line 8:
```suggestion
 - Domain
```

domain-model/README.md
```diff
@@ -0,0 +1,314 @@
+---
+layout: pattern
+title: Domain Model
+folder: domain-model
+permalink: /patterns/domain-model/
+categories: Structural
```
iluwatar, line 6:
```suggestion
categories: Architectural
```

iluwatar (changes requested):
I think it looks good already, but I'd like to change the category and tags for the pattern.

AndriyPyzh:
@iluwatar I committed your changes.

iluwatar (approved):

iluwatar:
Thank you for the new pattern!
@all-contributors please add @AndriyPyzh for code

PR #1811, opened 2021-08-21
Remove unnecessary and possibly not threadsafe flag

mortezaadi:
Remove unnecessary flag.

Introducing a new static flag to prevent instantiation via reflection is unnecessary. It's been defined as a static variable and therefore it is a shared data, it most likely needs proper synchronization.

besides, I believe it lacks proper readability.

Pull request description

- remove the flag
- check the instance variable to prevent reflection call


IAmPramod (approved):
Looks, good.
LGTM!

iluwatar (approved):

iluwatar:
Good reasoning, thank you for the improvement!
@all-contributors please add @mortezaadi for code

PR #1823, opened 2021-10-01
Update README.md to automatically add list of contributors

chandrikadeb7:
Pull request description

- Using dynamic web app [contributors-img](https://contrib.rocks/preview?repo=iluwatar%2Fjava-design-patterns)
- Reduces commits made for successive contributors
- Max Limit is set to 250, which can be altered as no. of contributors will increase in future


ohbus (commented):
Thanks for the fast PR.

But I am afraid that we need to discuss the tradeoffs with @iluwatar before proceeding with this.

chandrikadeb7:
@iluwatar Please suggest.

PR #1838, opened 2021-10-05
Refactor Prototype pattern

### LINKED ISSUE #584: Prototype pattern re-factor
Hi,

The Prototype implementations where clone() is Overridden (ElfBeast, ElfMage etc) can rather not have implementation of clone(). Instead their respective parent class (Mage, Beast) can have a general implementation of clone which creates a new instance by doing a shallow copy (Object.clone()) and then subsequently type-cast to the type of class (Mage.class or Beast.class etc) it is cloning for. 

This will have advantage of:

1. Removing verbosity among all implementations and dedicate the responsibility of cloning the object by their respective interfaces (abstract class).

2. Avoid using 'new' more than once and using only whilst cloning the prototype object that is passed in HeroFactoryImpl.

Regards,
panksy2k

ThexXTURBOXx:
This PR refactors parts of the Prototype pattern by making it `Cloneable` and thus inheriting the `clone()` method to its subclasses.

- Fixes #584

### DIFF
--- a/prototype/src/main/java/com/iluwatar/prototype/Beast.java
+++ b/prototype/src/main/java/com/iluwatar/prototype/Beast.java
@@ -37,6 +37,12 @@ public abstract class Beast implements Prototype {
   }
 
   @Override
-  public abstract Beast copy();
+  public Beast clone() {
+    try {
+      return (Beast) super.clone();
+    } catch (CloneNotSupportedException e) {
+      throw new RuntimeException(e);
+    }
+  }
 
 }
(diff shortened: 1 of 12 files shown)


yonatankarp (commented):
@iluwatar I can pick this one up ☝️

iluwatar:
Ok @yonatankarp

yonatankarp (commented):
There's open PR for this issue here: https://github.com/iluwatar/java-design-patterns/pull/1970 We can close this one...

iluwatar:
Ok @yonatankarp, I'll close this one

PR #1841, opened 2021-10-07
Add Indonesian translation

muklasr:
- Add README.md in localization/id/README.md
- Update README.md url


muklasr:
Thank you for approving my contribution!

PR #1847, opened 2021-10-11
Change some codes to make it thread-safe.

XinyuLiu5566:
- containsKey() and get() may not be thread-safe. In between the check and the get() another thread can remove the key and the get() will return null. calling get(), checking instead of your current check if the returned object is null, and then using that object only, without calling get() again.
- Replacing the check and put() with putIfAbsent() and then removing the synchronization. putIfAbsent() is thread-safe and puts the value only if the ConcurrentHashMap does not contain the key. putIfAbsent() returns null if the value did not exist and returns the value in the map if one already exists.


iluwatar:
Closed due to inactivity

PR #1866, opened 2021-10-17
Update Spacing of XML [Part 4]

ohbus:
Indentation Correction for XML files.

Housekeeping PR Set 4 of 4

- [x] reactor
- [x] reader-writer-lock
- [x] registry
- [x] repository
- [x] resource-acquisition-is-initialization
- [x] retry
- [x] role-object
- [x] saga
- [x] separated-interface
- [x] servant
- [x] serverless
- [x] service-layer
- [x] service-locator
- [x] sharding
- [x] singleton
- [x] spatial-partition
- [x] special-case
- [x] specification
- [x] state
- [x] step-builder
- [x] strangler
- [x] strategy
- [x] subclass-sandbox
- [x] table-module
- [x] template-method
- [x] thread-pool
- [x] throttling
- [x] tls
- [x] tolerant-reader
- [x] trampoline
- [x] transaction-script
- [x] twin
- [x] typeobjectpattern
- [x] unit-of-work
- [x] update-method
- [x] value-object
- [x] version-number
- [x] visitor

> Making the PR short so as to make it is easier to review.
> This PRs only deals with indentation.


iluwatar:
Thank you @ohbus for these formatting changes. It's good to have it consistent across the XML files. Additionally, I think it would be wise to investigate if we can enforce the correct indentation in the future. Perhaps Checkstyle can be configured to watch it?

PR #1915, opened 2021-11-21
Issue1483

AlexTamulaitis:
FIX: Issue 1483

Updates the Apache Isis version used for this pattern.


naked-objects/LICENSE
```diff
@@ -0,0 +1,201 @@
+ Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
```
iluwatar, line 3: Due to this license, we need to update the `LICENSE.md` at the root of the project to state that `naked-objects` is licensed under Apache

naked-objects/README.md
```diff
@@ -28,6 +28,10 @@ Use the Naked Objects pattern when
 
 * [Apache Isis](https://isis.apache.org/)
 
+## Reference
+
+* [Apache Isis SimpleApp](https://github.com/apache/isis-app-simpleapp)
+
```
iluwatar, line 34: Please add a section for an explanation that describes how the pattern works in practice. See the other patterns for examples of how it has been implemented

Additionally, it might be good to point out that `README` contains instructions on how to run this example.

iluwatar (changes requested):
Thank you for this enhacement! I added a couple of review comments that need to be addressed before merging. Additionally, I'm not sure why SonarCloud doesn't seem to be able to calculate the test coverage.

iluwatar:
Closed due to inactivity

PR #1928, opened 2021-12-12
Fix claim check frontmatter

iluwatar:
The new claim check pattern is not visible on java-design-patterns.com. I suspect the reason is that the pattern's frontmatter is missing the language definition. I'm adding it in this pull request.


ohbus (approved):
Thanks!

iluwatar:
Yes, I can confirm that this fixed the issue. Thanks @ohbus 👍🏻

Issue #1206, opened 2020-03-03
Event Aggregator pattern can be made more granular with emitting notification to specific observers


protyay:
The current implementation of the event -aggregator source code chooses to notify all the observers whenever an event source emits an event. 
This is can be made more granular with letting the subscriber subscribe with the aggregator for a specific type of event and then ONLY it should be notified in case , such an event is emitted. 

I would like to contribute a working version of these example, may be a PR if this issue is discussed.

iluwatar:
Sounds good @protyay :+1: I'll assign the issue to you. I'm looking forward to your pull request.

iluwatar:
@protyay are you still working on this?

protyay:
@iluwatar Hi Sorry, Got busy w/ my day job. I'm still interested but might need some time, especially with the pandemic around.

winiga-lab:
hey ! @iluwatar can you review my pr on this issue ;-) thanks

ohbus:
> hey ! @iluwatar can you review my pr on this issue ;-) thanks

Hi @winiga-lab, Please follow [Contributing Gudelines](https://github.com/iluwatar/java-design-patterns/wiki/01.-How-to-contribute)

winiga-lab:
hello @ohbus thanks for reply. can I change the implementation by the way. This implementation is more easier to understand https://www.youtube.com/watch?v=e6Cy6gf8EeY

ohbus:
> hello @ohbus thanks for reply. can I change the implementation by the way, [Events and Delegates: Event Aggregator](https://www.youtube.com/watch?v=e6Cy6gf8EeY) implementation is much easier to understand

Will be glad to do so! Please go on!

iluwatar:
@winiga-lab are you working on this?

iluwatar:
Ok @Leisterbecker

SriPrarabdha:
Hey , would love to work on this issue

iluwatar:
This issue was already fixed in https://github.com/iluwatar/java-design-patterns/pull/1933

PR #1933, opened 2021-12-21
EventAggregator now considers Event-Type 

Leisterbecker:
EventEmitter now contain a Map with Event as key and List<EventObserver> as value. Observer registration considers the event type and puts the observer into its special list within the map structure. Notification considers the event type and notifies only those who care for the event.
Changed some lines in the App.java class and added new Event: WHITE_WALKERS_SIGHTED.
The scout emits the new event at wednesday, Lord Varys observes and passes through.

Changed KingsHandTest to consider the new Event.

Related to #1206

Is the testing complete or should I add more cases related to the scout that emits a second event and varys that passes through?


iluwatar (approved):

iluwatar:
Good improvement, thank you!
@all-contributors please add @Leisterbecker for code

Issue #1948, opened 2022-01-15
Update Maven to the latest version


iluwatar:
As part of the Java 17 transition plan, the Maven version used to build the project should be updated to the latest. See https://github.com/iluwatar/java-design-patterns/wiki/19.-Java-17-transition

PR #1949, opened 2022-01-15
Update Maven

iluwatar:
Resolves issue https://github.com/iluwatar/java-design-patterns/issues/1948

- Updates Maven Wrapper and the Maven version it uses in the build
- Updates `maven-war-plugin` to the latest version to support the newer Maven versions


ohbus (approved):
LGTM

Thank you so much @iluwatar for keeping up with the housekeeping work!!

PR #1966, opened 2022-02-19
Add note about Project Lombok usage.

ErinsMatthew:
Add note about Project Lombok usage.

- Add a note about the fact that most of the solutions use Project Lombok.
- Allow users to more easily focus on design patterns.
- Help users know where to learn more about where to get Project Lombok for their IDE.


iluwatar:
Thanks for the suggestion, but I won't merge it

PR #1970, opened 2022-03-02
Refactor Prototype pattern (Closes #584)

### LINKED ISSUE #584: Prototype pattern re-factor

yonatankarp:
This PR refactors the Prototype pattern by making it `Cloneable` and thus inheriting the `clone()` method to its subclasses which removes code duplications.

* Fixes [Prototype pattern re-factor #584](https://github.com/iluwatar/java-design-patterns/issues/584)

### DIFF
--- a/lombok.config
+++ b/lombok.config
@@ -1,2 +1,2 @@
 lombok.log.fieldName = LOGGER
-
+lombok.addLombokGeneratedAnnotation = true
--- a/prototype/README.md
+++ b/prototype/README.md
@@ -22,7 +22,7 @@ used for creating new objects from prototype instances.
 
 Real-world example
 
-> Remember Dolly? The sheep that was cloned! Lets not get into the details but the key point here is 
+> Remember Dolly? The sheep that was cloned! Let's not get into the details but the key point here is 
 > that it is all about cloning.
 
 In plain words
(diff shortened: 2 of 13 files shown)


lombok.config
```diff
@@ -1,2 +1,2 @@
 lombok.log.fieldName = LOGGER
-
+lombok.addLombokGeneratedAnnotation = true
```
yonatankarp, line 2: Exclude Lombok auto-generated code from code-coverage

iluwatar (approved):

iluwatar:
This is a good refactoring that reduces the amount of boilerplate code. The code coverage is on a good level (75% on new code) even though Sonar is complaining.

iluwatar:
Thank you for the prototype pattern refactoring! 🎉

@all-contributors please add @yonatankarp for code

PR #1972, opened 2022-03-12
fix: a concurrent problem for lazy initialization

light0x00:
Pull request title

- use atomic operation `ConcurrentMap.computeIfAbsent` instead of double-check-lock  for lazy initialization

Pull request description

- fix the concurrent problem

### DIFF
--- a/reactor/src/main/java/com/iluwatar/reactor/framework/AbstractNioChannel.java
+++ b/reactor/src/main/java/com/iluwatar/reactor/framework/AbstractNioChannel.java
@@ -30,7 +30,7 @@ import java.nio.channels.Selector;
 import java.util.Map;
 import java.util.Queue;
 import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.ConcurrentLinkedQueue;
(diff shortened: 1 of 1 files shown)


ohbus (changes requested):
Please recheck why the builds have not been triggered and have failed.

Push your code again to retrigger the build.

algomaster99:
Hi! We are working on a tool that shows the "execution difference" of a patch. In other words, it reports if there has been a change in value of fields, local variables, or return. We tried to run our tool against the patch in this PR and got the following report (please open it a new tab to view it clearly):

Screenshot 2022-08-12 at 19-00-40 fix double lock check for lazy initialization · iluwatar_java-design-patterns@0e6045b


It would really help our research if the contributors could answer few of our questions.
1. Does the execution difference assist you in investigating the cause of the build failure?
2. Do you think it adds some more useful information that is not provided by the current state-of-the-art tools like Codecov, etc?

iluwatar:
@light0x00 can you describe the problem that the pull request is fixing? I don't see any failures in the master branch tests.

iluwatar:
Closed due to inactivity

PR #1856, opened 2021-10-15
Updated pom.xml (Dependency version of spring data jpa)

prakhyatsinghal:
I resolved the issue of a security alert in spring data jpa. #978  The dependency version was old, I updated it.

Issue #978, opened 2019-10-11
Security alert in spring-data-jpa


iluwatar:
Github reports security alert in spring-data-jpa dependency. It should be upgraded in this task.

![image](https://user-images.githubusercontent.com/582346/66677909-5603e200-ec73-11e9-961f-5d77bd1bc771.png)

Mistral:
Can I do this task?

iluwatar:
Go ahead @Mistral

bsshashank:
Sorry did not see that it was already assigned.

iluwatar:
@Mistral are you working on this?

prakhyatsinghal:
Can I work on this ?

iluwatar:
Yes, assigned to you @prakhyatsinghal, since @Mistral is not responding.

interactwithankush:
Please assign this to me.

mohanaraosv:
I have already done changes here https://github.com/iluwatar/java-design-patterns/pull/1870, Please assign to me.

iluwatar:
To be fair, I think @mohanaraosv expressed interest first. Let's try to complete this task in https://github.com/iluwatar/java-design-patterns/pull/1870

interactwithankush:
Sounds good @iluwatar

samratpodder:
What is the status of this Issue ? Is it resolved or is anyone working on it? @iluwatar

iluwatar:
It's not resolved yet. Would you like to pick it up @samratpodder?

akhil9281:
I would like to work on this issue for `hacktoberfest`.

pchugh17:
Let me know if contribution is needed for this. I would like to take this up or similar issues.

samratpodder:
> It's not resolved yet. Would you like to pick it up @samratpodder?

Yes, I would like to work on it. @iluwatar

iluwatar:
Assigned to @samratpodder

iluwatar:
@pchugh17 the issue is freed again if you want to take it?

iluwatar:
> samratpodder @samratpodder Oct 14 18:22
[...]

PR #1982, opened 2022-04-15
Upgraded spring-data-jpa and excluded overridden dependencies

mohanaraosv:
Inspired by the PR https://github.com/iluwatar/java-design-patterns/pull/1856

It resolves the issue of a security alert in spring data JPA. https://github.com/iluwatar/java-design-patterns/issues/978

I have picked this task to get familiar with the repo.
Upgraded the dependencies and excluded since version getting overridden from parent version.
Would like to upgrade the spring boot and config in the next PR.

### DIFF
--- a/repository/pom.xml
+++ b/repository/pom.xml
@@ -31,14 +31,28 @@
     <version>1.26.0-SNAPSHOT</version>
   </parent>
   <artifactId>repository</artifactId>
+  <properties>
+      <spring-data.version>2.6.3</spring-data.version>
+      <spring.version>5.3.19</spring.version>
+      <spring.boot.version>2.6.6</spring.boot.version>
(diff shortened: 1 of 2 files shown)


iluwatar:
Seems to work technically, but I think upgrading the spring-data dependency in the parent `pom.xml` (like #1856 does) is the right way to go. That's why I'm not going to merge this.

mohanaraosv:
> Seems to work technically, but I think upgrading the spring-data dependency in the parent `pom.xml` (like #1856 does) is the right way to go. That's why I'm not going to merge this.

I do agree, technically we should upgrade the main parent pom.  The split the unit of work I have followed this approach.   I will create another PR directly update in the main parent pom.

iluwatar:
Thanks @mohanaraosv, sounds good

Issue #1965, opened 2022-02-17
Translation error


liuyangyudaxia:
![image](https://user-images.githubusercontent.com/39300225/154443661-c12f821f-5942-4903-b158-8c5450711ab8.png)

xyllq999:
I want to try fixing this issue, thank u. @iluwatar

xyllq999:
Can you give more specific info about this bug or sth. ?

xmnh:
![image](https://user-images.githubusercontent.com/48951804/190574787-9a6ca2d2-522b-4aa0-80d8-60506ad54696.png)

iluwatar:
This issue is free for taking again.

nairpranav331:
Can you specify more about this.!

ohbus:
@nairpranav331 It is a Chinese translation error.

iluwatar:
Subtask of #2289

saiteja55:
I can fix this! Can I take it?

saiteja55:
@iluwatar can you please check this?
https://github.com/iluwatar/java-design-patterns/pull/2470

saiteja55:
#2470

saiteja55:
@iluwatar Can you check this please? 
#2470

PR #2005, opened 2022-05-24
fix a translation error

liuyangyudaxia:
- fix a translation error #1965


ohbus (changes requested):
According to Google Translate, I cannot find an error in this.

Issue #2033, opened 2022-09-28
Ensure plugin compatibility with Java 17


iluwatar:
This issue is a part of the Java 17 transition plan at https://github.com/iluwatar/java-design-patterns/wiki/19.-Java-17-transition.

The Maven build uses a plethora of plugins. In this issue let's ensure that all the used plugins have been updated to versions that support Java 17.

Acceptance criteria

- All the Maven plugins have been updated to recent versions that support Java 17

HeVictor:
Hi, I would like to give this a crack if that's alright. Thanks!

HeVictor:
Hi @iluwatar, I've been working on this issue and if it's alright I'd like to ask for some tips just to see if I'm on the right track. So far after forking and cloning the repo locally I've built the project in my IDE and I've configured it to use Java 17. Initially my approach was to go through each plugin to see if their versions have any release notes on whether it supports JDK 17 and then bump relevant ones in the POM file, but while some plugins mention this from their release notes I couldn't find this info for a lot of them anywhere (although sometimes I can find a couple of issue threads on other repos talking about fixing JDK compatibility issues for a plugin by bumping it above a certain version). 

Thus currently my approach is running `mvn clean install` to see what errors pop up from each of the design pattern directories and see if I can fix them through upping the plugin versions. My question is since there will be other incompatibility issues the project faces after switching to Java 17 as-is (such as due to library dependency incompatibilities), how should I make the changes in my PR to ensure plugin compatibility is met even if there's still errors during build or install? In other words what kind of testing steps should I take to verify that I've ensured compatibility for all the plugins in the project, beyond building the project in maven and running the unit tests?

I'm happy to put up a draft PR to show the changes I've made so far if it's easier to discuss where I'm at. Thanks so much for your time.

iluwatar:
I think executing `./mvnw clean install` should uncover 99% of the issues. Additionally, we can try to run each pattern using the command line e.g. `java -jar proxy.jar`, but the unit tests should mostly cover it already.

yonatankarp:
@HeVictor can we close this issue? or are there any other dependencies that should be migrated first?

HeVictor:
[...]

Issue #1683, opened 2021-03-18
Update the naked objects pattern example


iluwatar:
The naked objects example is based on the Apache Isis framework and currently already seems a bit outdated. In this task, we get the latest quickstart example (simpleapp) from https://isis.apache.org/ and fit it into `naked-objects` folder in our pattern catalog.

ramananrpn:
@iluwatar Shall I take this up ?

iluwatar:
Sure, go ahead @ramananrpn

Ramanan-Raguraman:
@iluwatar Could you please provide me write access ?

ramananrpn:
@iluwatar Could you please provide me write access ?

iluwatar:
This issue is free for taking again.

HattoriHenzo:
@iluwatar I would like to work on it, if still available.

iluwatar:
Yes, go ahead @HattoriHenzo

iluwatar:
@HattoriHenzo are you still working on this?

HattoriHenzo:
@iluwatar Sorry, I have started but not able to finish, because of lake of time. It seems there are many changes with the version of Apache Isis.

HeVictor:
@iluwatar I'm happy to give my crack at this as mentioned from my PR for #2033 if that's alright.

iluwatar:
Sure @HeVictor

Issue #1669, opened 2021-03-05
Tests not being executed in naked-objects


charlesfinley:
Naked objects module has tests that are being skipped during the maven build. 

These tests fail when executed by IntelliJ. They also fail if you try to update to JUnit5 via junit-vintage-engine. 

> This issue partially resolves #1500

charlesfinley:
To further clarify this issue - this module should be updated to JUnit 5 and the tests should be refactored as necessary.

PakhomovAlexander:
Hi @charlesfinley ! Could I take this?

charlesfinley:
> Hi @charlesfinley ! Could I take this?

Feel free. Remove any dependency on JUnit4 or junit-vintage-engine and have the tests pass.

ohbus:
> Hi @charlesfinley ! Could I take this?

Go ahead!

PakhomovAlexander:
Hi @charlesfinley @ohbus ! 

I've tried to migrate `naked-objects` module from junit 4 to junit 5. It went well for `dom` module. But it turned out that `integtests` module is deeply dependent on junit 4 rule API. It seems not possible to migrate without big changes. 

`naked-objects` pattern is a reference for apache isis. The current example is completely out of date. There is [simple-app](https://github.com/apache/isis-app-simpleapp) that uses junit 5. 

So, we can replace current `naked-objects` with `simple-app` example. Or just leave a reference to `simple-app` in order to have it always up to date.

charlesfinley:
> Hi @charlesfinley @ohbus !
> 
> I've tried to migrate `naked-objects` module from junit 4 to junit 5. It went well for `dom` module. But it turned out that `integtests` module is deeply dependent on junit 4 rule API. It seems not possible to migrate without big changes.
> 
> `naked-objects` pattern is a reference for apache isis. The current example is completely out of date. There is [simple-app](https://github.com/apache/isis-app-simpleapp) that uses junit 5.
> 
> So, we can replace current `naked-objects` with `simple-app` example. Or just leave a reference to `simple-app` in order to have it always up to date.

If the pattern is obsolete I don't see an issue getting rid of it, but that's obviously up to @iluwatar and @ohbus.

iluwatar:
[...]

PR #2051, opened 2022-10-04
Upgrade JaCoCo plugin for JDK 17 compatibility (#2033)

HeVictor:
This PR aims to solve #2033 by bumping JaCoCo plugin to 0.8.8, as attempting to run `mvn clean install` on the design modules under JDK 17 with the current version will result in the `java.lang.IllegalArgumentException: Unsupported class file major version 61` error.

For testing I've ran an overall `mvn clean install` on the entire project under JDK 17 with source and target also set to 17 in the `pom.xml`, and only the `naked-objects` pattern module produced an error (although there were warnings with other modules). It looks like the problem is rooted in an error running the `Datanucleus` enhancer inside `dom` (not much more error info) and errors inside `integtests` with using versions of the `Cucumber` library that's also too old from the Apache Isis dependency. I've tinkered with it a lot by updating dependencies and plugins inside `naked-objects` and still couldn't solve the errors, which is why I'm making this PR a draft.

I see there have already been discussions with `naked-objects` having some problems and being outdated from #1683 and #1669, and I'm happy to try rewrite the `naked-objects` to be more up-to-date and see if it can work with JDK 17 then. If it's crucial to have `naked-objects` build without errors under JDK 17 I'm happy to keep this PR in draft until the refactor is done.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -46,7 +46,7 @@
     <junit-jupiter.version>5.7.1</junit-jupiter.version>
     <junit-vintage.version>${junit-jupiter.version}</junit-vintage.version>
     <compiler.version>3.8.1</compiler.version>
-    <jacoco.version>0.8.6</jacoco.version>
+    <jacoco.version>0.8.8</jacoco.version>
     <commons-dbcp.version>1.4</commons-dbcp.version>
     <camel.version>2.24.0</camel.version>
     <guava.version>19.0</guava.version>


iluwatar (approved):

iluwatar:
Thanks @HeVictor for looking into this. I think this change is safe to merge as it doesn't break anything in master.

iluwatar:
@all-contributors please add @HeVictor for code

PR #2058, opened 2022-10-05
Update ZH document trans

wangchristine:
There are some english and chinese description both in zh documents. And one typo. This PR for fixing those.


wangchristine:
Why not accepted??

Issue #2062, opened 2022-10-07
add Filter pattern


MohammadRehanPatel:
please give PR with Hacktoberfest badge

skullkim:
I like to work on this issue.

MohammadRehanPatel:
I already have code for that woth a good example

iluwatar:
Ok, assigned to @MohammadRehanPatel. Can you also list an authoritative source for the pattern?

MohammadRehanPatel:
done

MohammadRehanPatel:
please add hacktoberfest tag

iluwatar:
The repository has the hacktoberfest tag. I believe it's enough for the contributors to submit any pull request, and once it gets approved, the hactoberfest PR count is increased.

MohammadRehanPatel:
I Learn this pattern during a course from udamy

aniketkumar-1907:
Can I work on this issue @skullkim.

drashan14:
Sir i would like to do work on this topic could you assign me this issue.

Zaid-Parkar:
Hello Sir,
I would like to contribute to this project,
I am new to hacktoctoberfest and this would be a great beginning.
Thank you...

ArchontisKostis:
is anyone assigned to this issue? Iwould like to work on it

MohammadRehanPatel:
I Already Done it

iluwatar:
The issue is free, if anyone wants to take a shot

lollanlollan:
We're a group of students that would like to work on this issue.

iluwatar:
Go ahead @lollanlollan

Pratik-kp:
Is this issue still open @iluwatar , I would like to take a shot

sugan0tech:
@Pratik-kp  can I know the status of your progress?

PR #2065, opened 2022-10-08
add filter pattern

skullkim:
Pull request title

- Add filter pattern
- This is to solve the issue #2062 

Pull request description

- Filter pattern intends to decouple objects and objects' filtering conditions through logical operations.


iluwatar (changes requested):
Please do not change the other patterns (data-mapper). Try to resolve the failing checks and then we can proceed with the review.

iluwatar:
Closed due to inactivity

Issue #1947, opened 2022-01-13
Chain Of Responsibilities Pattern Implementation using Java Interface/Implementation and Streams 


velmurugan0411:
Java Stream is simplifying of Implementing Chain of Responsibilities pattern. Interface with multiple implementations and run through chain via Java List streams.  This gives flexibility to switch the order of implementation chain without doing change. Can this be added to this pattern project?

For Eg.

interface ContactProcessor {
 Integer getPriority();
 boolean isApplicable() ;
  void process(Contact contact);
}

class Demographic implements ContactProcessor {

@Override
  public Integer getPriority() {
    return 1;
  }

 @Override
  public boolean isApplicable() {
    return true;
  }

 @Override
 void process(Contact contact) {
.....
 }
}

class Phone implements ContactProcessor {

@Override
  public Integer getPriority() {
    return 2;
  }

 @Override
  public boolean isApplicable() {
    return true;
  }

 @Override
 void process(Contact contact) {
.....
 }
}

class Email implements ContactProcessor {

@Override
  public Integer getPriority() {
    return 3;
  }

 @Override
  public boolean isApplicable(QualifiedBalContext context) {
    return true;
  }

 @Override
 void POA-444(Contact contact) {
.....
 }
}

Main Snippet....

@Autowired List<ContactProcessor > contactProcessor ;

contactProcessor .stream()
        .sorted(Comparator.comparing(ContactProcessor::getPriority))
        .filter(next -> next.isApplicable())
        .forEach(
            next -> {
             next.process(contact);
            });

iluwatar:
That's a nice trick. If someone wants to implement it by modifying the existing chain of responsibility example, I'll be happy to review the PR.

velmurugan0411:
Thanks for reviewing. I can pickup and provide Pull Request.

iluwatar:
Ok @velmurugan0411

perfectGay:
where Contact class?

iluwatar:
@velmurugan0411 are you still working on this?

JanFidor:
@iluwatar could I pickup this issue?

iluwatar:
I'm assuming that @JanFidor is still working on this

PR #2123, opened 2022-10-19
#1947 implement chain of responsibilities with java stream

JanFidor:
Use interface for RequestHandlers reduce code coupling. Logger could be injected with composition or chain could be sorted on creation, but I believe it would unnecessarily complicate the code.


iluwatar (approved):

iluwatar:
I like it @JanFidor. Thanks for this improvement!

@all-contributors please add @JanFidor for code

PR #2136, opened 2022-10-22
Add Nepali language localization support for the design patterns

0xYujan:
I just change in localization because there is no Nepali language so I just add the Nepali language. It helps Nepali people.


iluwatar (approved):

iluwatar:
Thank you for the new translation!

@all-contributors please add @0xYujan for translation

PR #2158, opened 2022-10-24
Fix #1929: Suppress logging hibernate internals

### LINKED ISSUE #1929: Hibernate of current version doesn't match h2
I use hibernate and h2 inherited from the parent pom.xml whose version are `5.2.18.Final`.
But I find the `org.hibernate.dialect.H2Dialect` extends from `org.hibernate.dialect.Dialect` didn't override method `getCurrentSchemaCommand`, which leeds it reture `null`.
Then I received an exception when I use hibernate to connect h2database:
```bash
org.hibernate.HibernateException: Use of DefaultSchemaNameResolver requires Dialect to provide the proper SQL statement/command but provided Dialect [org.hibernate.dialect.H2Dialect] did not return anything from Dialect#getCurrentSchemaCommand
	at org.hibernate.engine.jdbc.env.internal.DefaultSchemaNameResolver$SchemaNameResolverFallbackDelegate.resolveSchemaName(DefaultSchemaNameResolver.java:100)
	at org.hibernate.engine.jdbc.env.internal.DefaultSchemaNameResolver.resolveSchemaName(DefaultSchemaNameResolver.java:76)
......
```
It occurs becaus 
```bash
public String resolveSchemaName(Connection connection, Dialect dialect) throws SQLException {
	final String command = dialect.getCurrentSchemaCommand();
	if ( command == null ) {
		throw new HibernateException(
				"Use of DefaultSchemaNameResolver requires Dialect to provide the " +
						"proper SQL statement/command but provided Dialect [" +
						dialect.getClass().getName() + "] did not return anything " +
						"from Dialect#getCurrentSchemaCommand"
		);
	}
        ......
}
```
but the code can run successfully for some reasons. you can reproduce this case in my [code](https://github.com/castleKing1997/java-design-patterns/tree/master/metadata-mapping)
Is is necessary to fix it?
Maybe a new version of hibernate?

robertvolkmann:
I switched to log level info to suppress logging of hibernate internals by reusing the `logback.xml` from Maven project `cqrs`.

Fix #1929

### DIFF
--- a/metadata-mapping/README.md
+++ b/metadata-mapping/README.md
@@ -107,7 +107,7 @@ We use `Hibernate` to resolve the mapping and connect to our database, here's it
 <?xml version="1.0" encoding="utf-8"?>
 <!DOCTYPE hibernate-configuration PUBLIC
     "-//Hibernate/Hibernate Configuration DTD 3.0//EN"
-    "http://hibernate.sourceforge.net/hibernate-configuration-3.0.dtd">
+    "http://www.hibernate.org/dtd/hibernate-configuration-3.0.dtd">
 <hibernate-configuration>
   <session-factory>
     <!-- JDBC Database connection settings -->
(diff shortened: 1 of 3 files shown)


iluwatar (approved):

iluwatar:
The solution looks acceptable to me. Thanks for looking into this!

@all-contributors please add @robertvolkmann for code

Issue #1295, opened 2020-07-07
Embedded Value pattern


iluwatar:
https://www.youtube.com/watch?v=6n5O5mCdnew

shivu2002a:
Can I work on this issue?

iluwatar:
Assigned to @shivu2002a since he was the first. @kiratalreja3 I hope you find something else that interests you

shivu2002a:
@iluwatar Thank you for assigning me. What's the deadline to solve this issue?

iluwatar:
I'm not too strict, but if you would like to participate in Hacktoberfest, better do it during this month

shivu2002a:
Yeah sure. Thank you.

shivu2002a:
@iluwatar Hi. This pattern falls under the Relational-Structural design patterns but this category is not on the list of [categories and tags](https://github.com/iluwatar/java-design-patterns/wiki/07.-Categories-and-Tags), but there is a Structural pattern. I needed your suggestion regarding this and also about the tag that should be mentioned for this pattern. Thank you.

shivu2002a:
Need to be reviewed #2164

shivu2002a:
Need to be reviewed #2164

shivu2002a:
@iluwatar #2164 Need to be reviewed

iluwatar:
@shivu2002a, please address the review comments

shivu2002a:
@iluwatar I've made a commit saying 'adding suggested changes'.

PR #2164, opened 2022-10-25
Emdedded value design pattern

shivu2002a:
# Embedded Value Pattern

- Implementation of Embedded value pattern
- #1295 

## Description
- Implementation of Embedded value design pattern
- Complete pattern implementation along with class diagram and README 
- Implementation of test-cases
- Used H2 database
- I couldn't find enough resources for real-world examples and no Wikipedia page describing this pattern


embedded-value/src/main/java/com/iluwatar/embedded/value/DataSourceInterface.java
```diff
@@ -0,0 +1,69 @@
...
+import java.util.stream.Stream;
+
+/*
+ * Abstract class which contains the required SQL queries and basic methods declaration.
+ * 
+ * The main thing to consider is that the ShippingAddress object doesn't have it's own class
+ * but it's values are stored into the Orders table as city, state, pincode
+ * 
+ */
+interface DataSourceInterface {
```
robertvolkmann, line 37: Why do we need an interface? App.java does not even use it.

embedded-value/pom.xml
```diff
@@ -0,0 +1,66 @@
...
+  </parent>
+  <artifactId>embedded-value</artifactId>
+  <dependencies>
+    <dependency>
+      <groupId>com.h2database</groupId>
+      <artifactId>h2</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter-migrationsupport</artifactId>
```
robertvolkmann, line 43: Why do we need any migration support?
```suggestion
      <artifactId>junit-jupiter-engine</artifactId>
```

embedded-value/src/main/java/com/iluwatar/embedded/value/App.java
```diff
@@ -0,0 +1,85 @@
...
+  public static void main(String[] args) throws Exception {
+    final var dataSource = new DataSource();
+
+    final var order1 = new Order("JBL headphone", "Ram", new ShippingAddress("Bangalore", "Karnataka", "560040"));
+    final var order2 = new Order("MacBook Pro", "Manjunath",
+        new ShippingAddress("Bangalore", "Karnataka", "581204"));
+    final var order3 = new Order("Carrie Soto is Back", "Shiva",
+        new ShippingAddress("Bangalore", "Karnataka", "560004"));
+    
+        if (dataSource.createSchema()) {
```
robertvolkmann, line 63:
```suggestion
    if (dataSource.createSchema()) {
```

embedded-value/src/main/java/com/iluwatar/embedded/value/Order.java
```diff
@@ -0,0 +1,63 @@
...
+import lombok.NoArgsConstructor;
+import lombok.Setter;
+import lombok.ToString;
+/* 
+ * A POJO which represents the Order object.
+ */
+@ToString
+@Setter
+@Getter
+@NoArgsConstructor
```
robertvolkmann, line 37: Not used.
```suggestion
```

embedded-value/src/main/java/com/iluwatar/embedded/value/ShippingAddress.java
```diff
@@ -0,0 +1,58 @@
...
+import lombok.Setter;
+import lombok.ToString;
+
+/** 
+ * Another POJO which wraps the Shipping details of order into Object.
+ */
+@ToString
+@Setter
+@Getter
+@NoArgsConstructor
```
robertvolkmann, line 38: Not used.
```suggestion
```

embedded-value/src/main/java/com/iluwatar/embedded/value/DataSource.java
```diff
@@ -0,0 +1,217 @@
...
+
+  @Override
+  public Stream<Order> queryOrders() {
+    var ordersList = new ArrayList<Order>();
+    try (var rSet = queryOrders.executeQuery(QUERY_ORDERS)) {
+      while (rSet.next()) {
+        var order = new Order(rSet.getString(2), rSet.getString(3),
+            new ShippingAddress(rSet.getString(4), rSet.getString(5),
+            rSet.getString(6)));
+        order.setId(rSet.getInt(1));
```
robertvolkmann, line 150: Why not using the constructor of Order to supplier the id instead of using a Setter?

embedded-value/src/main/java/com/iluwatar/embedded/value/DataSource.java
```diff
@@ -0,0 +1,217 @@
...
+      insertIntoOrders.setString(2, order.getOrderedBy());
+      var address = order.getShippingAddress();
+      insertIntoOrders.setString(3, address.getCity());
+      insertIntoOrders.setString(4, address.getState());
+      insertIntoOrders.setString(5, address.getPincode());
+      var affectedRows = insertIntoOrders.executeUpdate();
+      if (affectedRows == 1) {
+        var rs = insertIntoOrders.getGeneratedKeys();
+        rs.last();
+        order.setId(rs.getInt(1));
```
robertvolkmann, line 124: Why you are mutating the parameter instead of returning a new one?

embedded-value/src/main/java/com/iluwatar/embedded/value/ShippingAddress.java
```diff
@@ -0,0 +1,58 @@
...
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+import lombok.Setter;
+import lombok.ToString;
+
+/** 
+ * Another POJO which wraps the Shipping details of order into Object.
+ */
+@ToString
+@Setter
```
robertvolkmann, line 36: Not used.
```suggestion
```

embedded-value/src/main/java/com/iluwatar/embedded/value/Order.java
```diff
@@ -0,0 +1,63 @@
...
+
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+import lombok.Setter;
+import lombok.ToString;
+/* 
+ * A POJO which represents the Order object.
+ */
+@ToString
+@Setter
```
robertvolkmann, line 35: After adjusting DataSource we don't need any Setter.
```suggestion
```

robertvolkmann (commented):
What is your opinion about extracting the Order stuff into a separate Service to follow the single responsibility principle?

embedded-value/src/main/java/com/iluwatar/embedded/value/DataSourceInterface.java
shivu2002a: > What is your opinion about extracting the Order stuff into a separate Service to follow the single responsibility principle?
Sorry, I didn't get it. Can you explain a bit?


> Why do we need an interface? App.java does not even use it.
I thought it'd be appropriate to keep the SQL queries and methods in one place.

embedded-value/pom.xml
shivu2002a: > Why do we need any migration support?

Yeah, we don't need it in this case.

embedded-value/src/main/java/com/iluwatar/embedded/value/DataSource.java
shivu2002a: I'll return a new one then.

embedded-value/src/main/java/com/iluwatar/embedded/value/Order.java
shivu2002a: Should I commit these changes using the 'Commit suggestion' button below the suggested changes or by modifying the code on my laptop and then commit?

robertvolkmann: It doesn't matter. Both works.

robertvolkmann:
@shivu2002a The whole JDBC and H2 stuff complicates this simple pattern. What about using a custom inmemory database, that is only a wrapper around an ArrayList?

@iluwatar What's your opinion?

```
public class OrderStore {
    private final ArrayList<OrderItem> store;
    
    public OrderStore() {
        store = new ArrayList<>();
    }
    
    public boolean insert(Order order) {
        if (store.add(order)) {
            return new Order(store.size() - 1);
        };
    }
    
    public List<Order> getAll() {
        return store.stream()
            .filter(Objects::nonNull)
            .map( /* map to Order */ )
            .collect(Collectors.toList());
    }
    
    public void remove(int id) {
        store[id] = null;
    }
    
    private static class OrderItem {
        /* properties */
    } 
}
```

shivu2002a:
@robertvolkmann should I implement this using an in-memory database?

embedded-value/src/main/java/com/iluwatar/embedded/value/App.java
```diff
@@ -0,0 +1,85 @@
...
+    LOGGER.info("Orders Query: {}", dataSource.queryOrders().collect(Collectors.toList()));
+    LOGGER.info("Query Order with id=2: {}", dataSource.queryOrder(2));
+    LOGGER.info("Remove Order with id=1");
+    dataSource.removeOrder(1);
+    LOGGER.info("Orders Query: {}", dataSource.queryOrders().collect(Collectors.toList()));
+    if (dataSource.deleteSchema()) {
+      LOGGER.info("TABLE DROPPED");
+    } else {
+      LOGGER.error("Error deleting table");
+    }
```
iluwatar, line 83: Please add some strategic comments that explain what's going on. Keep in mind that this example is for other people to study and learn from.

embedded-value/src/main/java/com/iluwatar/embedded/value/DataSource.java
```diff
@@ -0,0 +1,217 @@
...
+    try {
+      createschema = conn.createStatement();
+      createschema.execute(CREATE_SCHEMA);
+      insertIntoOrders = conn.prepareStatement(INSERT_ORDER, PreparedStatement.RETURN_GENERATED_KEYS);
+      getschema = conn.createStatement();
+      queryOrders = conn.createStatement();
+      removeorder = conn.prepareStatement(REMOVE_ORDER);
+      queyOrderByID = conn.prepareStatement(QUERY_ORDER);
+      deleteschema = conn.createStatement();
+      createschema.close();
```
iluwatar, line 86: You can use try-with-resource to autoclose resources

iluwatar: Check this elsewhere too

iluwatar (changes requested):

iluwatar:
> @shivu2002a The whole JDBC and H2 stuff complicates this simple pattern. What about using a custom inmemory database, that is only a wrapper around an ArrayList?
> 
> @iluwatar What's your opinion?
> 
> Sketch:
> 
> ```
> public class OrderStore {
>     private final ArrayList<OrderItem> store;
>     
>     public OrderStore() {
>         store = new ArrayList<>();
>     }
>     
>     public Order insert(Order order) {
>         var item = new OrderItem( /* values from order */ )
>         if (store.add(item)) {
>             return new Order(store.size() - 1, /* other values from order */);
>         }
>         return null;
>     }
>     
>     public List<Order> getAll() {
>         return store.stream()
>             .filter(Objects::nonNull)
>             .map( /* map to Order */ )
>             .collect(Collectors.toList());
>     }
>     
>     public void remove(int id) {
>         store[id] = null;
>     }
>     
>     private static class OrderItem {
>         /* properties */
>     } 
> }
> ```

We can do it either way. You are right that this would make the code simpler but on the other this is a data access pattern and using a real(ish) database is justified.

embedded-value/pom.xml
```diff
@@ -0,0 +1,66 @@
...
+    THE SOFTWARE.
+
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>com.iluwatar</groupId>
+    <artifactId>java-design-patterns</artifactId>
+    <version>1.26.0-SNAPSHOT</version>
+  </parent>
```
iluwatar, line 34: You also need to add this new pattern as a module in the parent pom.xml. Otherwise, it won't be built at all.

iluwatar (changes requested):
Please rebase this branch on top of `master` and address any issues that come from the build.

embedded-value/pom.xml
shivu2002a: Yeah, sure.

Issue #2151, opened 2022-10-23
Add Model-View-Intent design pattern


JanFidor:
I noticed that while MVP, MVC and MVVM have been added, it wasn't the case for MVI. I'll be happy fix that.

iluwatar:
Thanks @JanFidor for the suggestion. Could you provide a link that explains the pattern?

JanFidor:
Sure thing, here are 2 of the best ones I found https://apiumacademy.com/blog/model-view-intent-pattern/  , https://www.raywenderlich.com/817602-mvi-architecture-for-android-tutorial-getting-started  . The pattern was created with android in mind, so most explanations lean heavily on it with explanations. It also heavily utilizes sealed classes from kotlin, but it can still be implemented in Java and I think it's worth a mention for its focus on encapsulation.

iluwatar:
Looks great! Thanks so much for providing the links.

iluwatar:
I assume @JanFidor is still working on it

PR #2177, opened 2022-10-26
#2151 Feature/mvi pattern

JanFidor:
Add an MVI design pattern (includes pattern itself, tests, README.md and class diagram)


iluwatar:
The build is failing. Please investigate @JanFidor

JanFidor:
@iluwatar It seems an independent module is causing issues (page-object). My build is failing even on the master branch after syncing the fork. The first error is caused by missing test dependencies, after adding them [this line ](https://github.com/iluwatar/java-design-patterns/blob/e6f5b691df3033ad3a73c14f6ad9671efa49bb03/page-object/src/test/java/com/iluwatar/pageobject/pages/AlbumListPage.java#L82) throws an error `java: incompatible types: java.util.List<java.lang.Object> cannot be converted to java.util.List<com.gargoylesoftware.htmlunit.html.HtmlAnchor>`
I'm not certain if I managed to break git on my part or if it's something else.

iluwatar:
The error from the CI build is `Failed to execute goal org.apache.maven.plugins:maven-checkstyle-plugin:3.1.0:check (validate) on project model-view-intent: You have 2 Checkstyle violations.` @JanFidor

JanFidor:
@iluwatar  my bad for misinterpreting logs and sorry for the delay, university gave me plenty of work this week, but I'll get right on it now.

JanFidor:
@iluwatar fixed the violations, was a little closer to 60 after I run checkstyle locally, but it checks out now

model-view-intent/README.md
```diff
@@ -0,0 +1,47 @@
...
+---
+
+## Intent
+MVI is a derivation of the original MVC architectural pattern. Instead of working with a 
+proactive controller MVI works with the reactive component called intent: it's a component
+which translates user input events into model updates.
+
+## Class diagram
+![alt text](./etc/model-view-intent.png "Model-View-Intent")
+
```
iluwatar, line 17: We are missing explanation and programmatic example sections

model-view-intent/src/main/java/com/iluwatar/model/view/intent/App.java
```diff
@@ -0,0 +1,66 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package com.iluwatar.model.view.intent;
+
+/**
+ * Model-View-Intent is a pattern for implementing user interfaces.
+ */
```
iluwatar, line 31: In this section, please explain the pattern briefly and describe how this example implements it

model-view-intent/src/main/java/com/iluwatar/model/view/intent/App.java
```diff
@@ -0,0 +1,66 @@
...
+    view.setVariable(variable1);
+    view.add();
+    view.displayTotal();
+
+    variable1 = 2.0;
+    view.setVariable(variable1);
+    view.subtract();
+    view.divide();
+    view.multiply();
+    view.displayTotal();
```
iluwatar, line 58: - Use `var` for local variables, where possible
- Add more comments for the audience (this is study material)

model-view-intent/src/main/java/com/iluwatar/model/view/intent/CalculatorModel.java
```diff
@@ -0,0 +1,61 @@
...
+  /**
+   * Create calculator model with starting values.
+   *
+   * @param newOutput   -> starting output value.
+   * @param newVariable -> starting variable value.
+   **/
+  public CalculatorModel(final Double newOutput, final Double newVariable) {
+    this.output = newOutput;
+    this.variable = newVariable;
+  }
```
iluwatar, line 27: Use Lombok to reduce boilerplate

iluwatar: Check this throughout the code

model-view-intent/src/main/java/com/iluwatar/model/view/intent/CalculatorView.java
```diff
@@ -0,0 +1,70 @@
...
+ * {@link com.iluwatar.model.view.intent.actions.CalculatorAction}
+ * and displays its updated {@link CalculatorModel}.
+ */
+@Slf4j
+public class CalculatorView {
+
+  /**
+   * View model param handling the operations.
+   * */
+  private final CalculatorViewModel viewModel = new CalculatorViewModel();
```
iluwatar, line 22: Usually it's better to pass variables through constructors or methods to avoid tight coupling to single implementation

model-view-intent/src/main/java/com/iluwatar/model/view/intent/actions/CalculatorAction.java
```diff
@@ -0,0 +1,15 @@
+package com.iluwatar.model.view.intent.actions;
+
+/**
+ * Defines what outside interactions can be consumed by view model.
+ * */
+public interface CalculatorAction {
```
iluwatar, line 6: I think you can use an enum-based implementation to get rid of the zillion little `CalculatorAction` derived classes

iluwatar (changes requested):

JanFidor:
Hi, @iluwatar  sorry for the long break, but the past couple of weeks have been pretty crazy on my end, but I'll be getting on it now

model-view-intent/src/main/java/com/iluwatar/model/view/intent/actions/CalculatorAction.java
JanFidor: I thought about it, but the main strength of the MVI pattern is the reduced API of the ViewModel. The only way to achieve it in a non brittle way is having an option to pass different amounts of arguments with different types to an Intent. In Kotlin it's straight thanks to sealed classes, as far as I'm aware, Java Enums don't support constructors with differing parameters, which would both significantly limit the use cases and hide MVIs main feature.

iluwatar:
Lift doesn't seem to handle Lombok annotations too well, so I think we can disregard them.

iluwatar (approved):

iluwatar:
Looks good! Thank you for contributing this new pattern 🎉

@all-contributors please add @JanFidor for code

PR #2190, opened 2022-10-28
Localize active-object to Vietnamese

m-inh:
I translate `active-object` pattern to Vietnamese


iluwatar (approved):

iluwatar:
Another one translated, thank you!

Issue #1219, opened 2020-04-04
Rule engine design pattern 


mohanaraosv:
**Description:**
The Rules design pattern is a behavioral design pattern that enables the encapsulation of business rules into isolated, reusable units. This pattern allows for the dynamic application of business logic at runtime and promotes the separation of concerns, making the system more modular and easier to maintain.

**Main Elements of the Pattern:**
1. **Rule Interface**: Defines the contract for all rules, typically including a method to evaluate the rule and another to execute the rule's action.
2. **Concrete Rules**: Implementations of the Rule interface, each encapsulating a specific piece of business logic.
3. **Rule Engine**: Responsible for managing and executing rules. It evaluates which rules should be applied based on the given context and executes them accordingly.
4. **Context**: The data or state that the rules operate on. This is passed to the rule engine and rules during execution.

**References:**
- [Rules Design Pattern by Michael Whelan](https://www.michael-whelan.net/rules-design-pattern/)
- [Project Contribution Guidelines](https://github.com/iluwatar/java-design-patterns/wiki)

**Acceptance Criteria:**
1. Create a Rule interface with methods to evaluate and execute a rule.
2. Implement at least three concrete rule classes demonstrating different business logic scenarios.
3. Develop a RuleEngine class capable of evaluating and executing a set of rules against a provided context.
4. Include unit tests for the RuleEngine and the concrete rule classes to ensure correctness and robustness.

mohanaraosv:
I will work on this.

iluwatar:
Great @mohanaraosv

iluwatar:
@mohanaraosv are you working on it?

mohanaraosv:
I didn't get a chance to work on this, But I would like to contribute to this let it be on my name.

iluwatar:
Ok @mohanaraosv, let me know if you require assistance

arjunduggal:
hi @iluwatar 
Can I pick this up if this is still open?
Thanks

iluwatar:
@mohanaraosv was assigned earlier, but no news yet. Are you still working on it @mohanaraosv or could @arjunduggal take it?

mohanaraosv:
Sorry for the delay!! I have made some progress on this please pick @arjunduggal  some other pattern.

guapi777:
I would love to solve this issue within one month.........

ohbus:
Thanks for your interest @guapi777 

[...]

PR #2191, opened 2022-10-28
Issue#1219 Rule Engine Design Pattern Implemented

u7076185:
* Resolves issue Rule Engine Design Pattern #1219 

* The idea of the pattern looks as follows:
The Rules Pattern works by separating the rules from the rules processing logic (applying the Single Responsibility Principle). This makes it easy to add new rules without changing the rest of the system (applying the Open/Closed Principle).
With this in mind, we implement the pattern by having an imaginary country that has an immigration policy with multiple criteria. We create a separate class for each criterion to implement the pattern to calculate the candidate's immigration score.

* A directory called Rule Engine Design Pattern has been added to present the implementation of the pattern, built based on the instruction https://github.com/iluwatar/java-design-patterns/wiki/01.-How-to-contribute.


rule-engine/README.md
```diff
@@ -0,0 +1,32 @@
...
+![image](https://user-images.githubusercontent.com/110162648/198545044-4d45db23-0ed9-4369-881b-2b4261410222.png)
+
+# Wikipedia says
+Rules design pattern helps the developer to encapsulate each business rule in a separate object and decouple the definition of business rules from their processing. New rules can be added without the need to modify the rest of the application logic.
+## Class Diagram
+
+![image](https://user-images.githubusercontent.com/110162648/198545171-a154750a-e6a9-4785-b183-b97303e7ff8a.png)
+
+
+
```
iluwatar, line 32: Regarding the README.md implementation, the general rules are specified in https://github.com/iluwatar/java-design-patterns/wiki/01.-How-to-contribute

- Adding the frontmatter yaml is needed
- Explanation should follow the same format as the other patterns. See https://github.com/iluwatar/java-design-patterns/wiki/02.-Pattern-template and for example https://github.com/iluwatar/java-design-patterns/blob/master/bridge/README.md
- Programmatic example should be copyable text, not images

rule-engine/pom.xml
```diff
@@ -0,0 +1,59 @@
...
+            <artifactId>junit-jupiter-api</artifactId>
+            <scope>test</scope>
+        </dependency>
+    </dependencies>
+
+    <properties>
+        <maven.compiler.source>18</maven.compiler.source>
+        <maven.compiler.target>18</maven.compiler.target>
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    </properties>
```
iluwatar, line 57: Please use the Java version from the parent pom.xml

rule-engine/pom.xml
```diff
@@ -0,0 +1,59 @@
...
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>rule-engine</artifactId>
+    <dependencies>
+        <dependency>
+            <groupId>org.testng</groupId>
+            <artifactId>testng</artifactId>
+            <version>RELEASE</version>
+            <scope>test</scope>
+        </dependency>
```
iluwatar, line 45: In general, we use JUnit for unit testing

rule-engine/src/main/java/com/iluwatar/ruleengine/App.java
```diff
@@ -0,0 +1,81 @@
...
+ * immigration score.
+ */
+
+@Slf4j
+public class App {
+  /**
+   * The main method of rule engine pattern.
+   */
+  public static void main(String[] args) {
+    List<ImmigrationRule> rules = new ArrayList<>();
```
iluwatar, line 49: Use `var` keyword for local variables where possible

rule-engine/src/main/java/com/iluwatar/ruleengine/App.java
```diff
@@ -0,0 +1,81 @@
...
+      if (rule.shouldRun(candidate1)) {
+        score1 += rule.runRule(candidate1);
+        score2 += rule.runRule(candidate2);
+        score3 += rule.runRule(candidate3);
+      }
+    }
+
+    LOGGER.info("score for candidate 1: " + score1);
+    LOGGER.info("score for candidate 2: " + score2);
+    LOGGER.info("score for candidate 3: " + score3);
```
iluwatar, line 78: Please explain by using comments what is happening here

pom.xml
```diff
@@ -80,6 +80,7 @@
     <sonar.projectKey>iluwatar_java-design-patterns</sonar.projectKey>
     <sonar.moduleKey>${project.artifactId}</sonar.moduleKey>
     <sonar.projectName>Java Design Patterns</sonar.projectName>
+      <kotlin.version>1.5.10</kotlin.version>
```
iluwatar, line 83: Hmm, could you explain why we need Kotlin?

iluwatar (changes requested):

PR #2192, opened 2022-10-28
Refactor REAME.md in `/localization/vi` folder

m-inh:
The files under `localization` folder are not supposed to change frequently, so it seems redundant to keep some content that generated/injected by a bot. Sequentially, it will become outdated/false by time. User can take this kind of information in the original `README.md`.


iluwatar (approved):

iluwatar:
I agree with you @m-inh 

@all-contributors please add @m-inh for translation

Issue #1483, opened 2020-08-08
Update/replace Naked Objects example


iluwatar:
Naked Objects example has been implemented using a very old Apache Isis version. In this task we'll replace it with an example generated using the latest Isis.

Apache Isis framework: https://isis.apache.org/

dieuph:
Hi @iluwatar, I'll take it up.

iluwatar:
Thanks @dieuph, go ahead!

berlincho:
@dieuph, did you fix the issue? Thanks.

ohbus:
Thank you for your efforts!

berlincho:
@ohbus Although the current example uses the old version of Apace isis, could you share the info of how to run it?

iluwatar:
Good question. Previously we could follow the instructions at https://github.com/iluwatar/java-design-patterns/blob/master/naked-objects/README, but it seems that they removed it from the website.

berlincho:
@iluwatar Thanks for your reply. I check the webpage archive (http://web.archive.org/web/20150717233428/http://isis.apache.org/guides/ug.html#3.-getting-started). 
It seems that the existing example uses the sample application in the above link. I will replace it with the newer Apache isis version.

iluwatar:
Great @berlincho, I'm looking forward to the PR

chirag-cyber:
hi @iluwatar can you this task to me?

iluwatar:
Thanks @chirag-cyber, this issue is much anticipated

Issue #1097, opened 2019-11-17
"Unknown host repository-estatio.forge.cloudbees.com" issue


numbernine9:
Hi, this is Masaki Mizobuchi.
I faced the below error when I build the projects. So, I report is.

**Unknown host repository-estatio.forge.cloudbees.com** is shown when I execute the commnad `mvn eclipse:eclipse`.
It seems that [(http://repository-estatio.forge.cloudbees.com/snapshot/)] is closed host.

Error message is below.

> [ERROR] Failed to execute goal on project naked-objects-fixture: Could not resolve dependencies for project com.iluwatar:naked-objects-fixture:jar:1.22.0-SNAPSHOT: Could not transfer artifact com.iluwatar:naked-objects-dom:jar:1.22.0-SNAPSHOT from/to Cloudbees snapshots (http://repository-estatio.forge.cloudbees.com/snapshot/): repository-estatio.forge.cloudbees.com: Unknown host repository-estatio.forge.cloudbees.com -

iluwatar:
Confirmed, I can repeat this.

PR #2336, opened 2022-11-15
feat: remove naked objects implementation

### LINKED ISSUE #983: Java 11 support: Naked Objects pattern
In issue https://github.com/iluwatar/java-design-patterns/issues/976 JDK11 build was taken to use but one problem related to Naked Objects pattern remains. Issue description from @lbroman:

> In my PR #980 I have left language source and target at 8 for the naked-objects/dom module. This is because java 11 bytecode breaks the datanucleus enhancer, even if they claim that 5.2 should support java 11 bytecode.
I have tried fiddling with the datanucleus dependency versions because the module uses datanucleus core 5 and apis from 4 so I thought that might be the problem. However, bumping up the versions breaks other stuff as they seem to have moved around some classes between 4 and 5.

robertvolkmann:
Added link to the Apache Isis starter SimpleApp.

Fixes #983 #1097 #1483 #1669 #1683


iluwatar (approved):

iluwatar:
Wow, that certainly fixed a lot of issues. Thank @robertvolkmann for solving this puzzle! 😄

Issue #2362, opened 2022-11-26
Extend naked-objects README


HeVictor:
Due to the decision to remove the naked-objects implementation as a part of #2336, the work that I had been doing for #1683 has become largely obsolete but I would still like to contribute the extensions I have made to the README doc itself.

PR #2363, opened 2022-11-26
docs: extend naked-objects README (#2362)

HeVictor:
Extends the naked-objects README to describe the pattern and its uses in more detail. 

Special note is that I propose using a `Quick start` heading to provide the link to the _Apache Isis_ starter app instead of `Real world examples` as I think it's clearer to show that this is where readers can go and see how this pattern really works, as it is unique from the other patterns because it needs a framework implementation of the pattern itself before one can interact with it.


iluwatar (approved):

iluwatar:

@all-contributors please add @HeVictor for docs

PR #2464, opened 2023-01-28
#2449 bump maven-checkstyle-plugin from 3.1.0 to 3.2.0 + resolve chec…

rahul-raj:
The intent of this PR is tp upgrade the checkstyle version from 3.1.0 to 3.2.0 and resolve all checkstyle issues.

**Checkstyle issues resolved ->**

- Multiple classes without JavaDoc comment in it.
- Class files, Method names, variable names with consecutive caps. Renamed and refactored them accordingly. 
- Line space before attributes (such as @param) in the JavaDoc comment section.
- Remove linespaces between import statements.
- Adding space (' ') between semicolons in the loop, bring logical operators (such as &&) in new lines. Checkstyles related to formatting in general.

**Note ->**
Added `<suppress checks="RequireEmptyLineBeforeBlockTagGroup" files=".*\.java"/>` in the suppressions. 
This is to ignore line space requirement on the JavaDoc section. 

There are two reasons behind this new suppression rule:
1) Need to change 300+ files for this minor checkstyle condition.
2) When we press "enter" while in JavaDoc comment section, all IDEs (tested on Vscode & IntelliJ) insert a space and then goes to next line. So, maven think that it is on the same line and will result in checkstyle error. This will create some confusion to people while resolving the checkstyle issues. 

For me, this is not super important and unnecessary . But, if you think, it is required, then let me know. I will then go ahead and make those changes. 

Sorry about the multiple commits as there was an existing `etc/` directory which was removed in the checkin and I added it back. Also, there was a  file `FileSelectorJFrame.java`  that was renamed to `FileSelectorJframe.java`  in favor of checkstyle. But it wasn't checked-in for some strange reason. I fixed that as well.


iluwatar (approved):

iluwatar:

@all-contributors please add @rahul-raj for code

PR #2503, opened 2023-04-04
Code Refactoring

sourav15102:
### Pull request title: Refactoring enhancements

I have made some refactoring changes in the code fora few design patterns.
It makes the code more readable and more cohesive.

### Pull request description

- Design pattern: Caching
- Refactoring method: Push down method
- Changes:
> 1. CacheStore contained a lot of methods which can be pushed down to child classes.
> 2. To make it more cohesive, I divided the class into two more classes: CacheStoreRead, CacheStoreWrite.
> 3. This way the code can be considered more extensible.

- Design patter: Commander
- Refactoring method: Decompose conditional
- Changes:
> 1. The if conditional added complexity to the already complex method.
> 2. I took out the if condition and put it in another method, reducing the complexity a little bit.

- Design pattern: Domainmodel
- Refactoring method: Extract class
- Changes:
> 1. There were a lot of responsibilities handled by the class 'Customer'.
> 2. I made a separate class called 'Account' which will handle all the customer's account details.
> 3. Class 'Customer' contains its 'Account'.

- Design pattern: intercepting-filter
- Refactoring method:pull-up method
- Changes:
> 1. All the child classes of 'Abstract Filter' implements execute method, and they are following a pattern.
> 2. We can create a template of this function in 'Abstract Filter', and force child classes to implemented helper method used in the template.

- Design pattern: typeobject
- Refactoring method: extract method
- Changes:
> 1. Some blocks of code in method 'round' in class 'CandyGame' can be taken out.
> 2. Three more functions were created to reduce complexity of function.
> 3. Changes will make it more testable.

- Design pattern: balking
- Refactoring method: Introducing explaining variable
- Changes:
> 1. The variable with value '50' was directly used.
> 2. Introduced variable with explaining names.

### Testing
- There was one test which was failing at the time i forked the repository.
> Design pattern: dirty-flag
- All the other tests are passing.

Please let me know if I need to make any further changes.

### DIFF
--- a/caching/src/main/java/com/iluwatar/caching/AppManager.java
+++ b/caching/src/main/java/com/iluwatar/caching/AppManager.java
@@ -50,6 +50,8 @@ public class AppManager {
    * Cache Store.
    */
   private final CacheStore cacheStore;
+  private final CacheStoreRead cacheStoreRead;
+  private final CacheStoreWrite cacheStoreWrite;
 
   /**
    * Constructor.
@@ -59,6 +61,8 @@ public class AppManager {
   public AppManager(final DbManager newDbManager) {
     this.dbManager = newDbManager;
     this.cacheStore = new CacheStore(newDbManager);
+    this.cacheStoreRead = new CacheStoreRead(newDbManager);
+    this.cacheStoreWrite = new CacheStoreWrite(newDbManager);
   }
 
   /**
@@ -93,9 +97,9 @@ public class AppManager {
     LOGGER.info("Trying to find {} in cache", userId);
     if (cachingPolicy == CachingPolicy.THROUGH
             || cachingPolicy == CachingPolicy.AROUND) {
-      return cacheStore.readThrough(userId);
+      return cacheStoreRead.readThrough(userId);
     } else if (cachingPolicy == CachingPolicy.BEHIND) {
-      return cacheStore.readThroughWithWriteBackPolicy(userId);
+      return cacheStoreRead.readThroughWithWriteBackPolicy(userId);
     } else if (cachingPolicy == CachingPolicy.ASIDE) {
       return findAside(userId);
     }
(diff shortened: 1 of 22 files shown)


yoavami (approved):
Changes look good.

caching/src/main/java/com/iluwatar/caching/AppManager.java
```diff
@@ -50,6 +50,8 @@ public class AppManager {
    * Cache Store.
    */
   private final CacheStore cacheStore;
+  private final CacheStoreRead cacheStoreRead;
+  private final CacheStoreWrite cacheStoreWrite;
```
iluwatar, line 54: It's a good inititiative to refactor this, but I find it confusing that we now have three different cache classes and objects.

iluwatar: One option would be to create a cache class for each of the use cases we want to demonstrate (write-through, write-around, write-behind, cache-aside, read-through)

commander/src/main/java/com/iluwatar/commander/Commander.java
```diff
@@ -245,12 +245,7 @@ private void updateQueue(QueueTask qt) {
       // additional check not needed
       LOG.trace(ORDER_ID + ": Queue time for order over, failed..", qt.order.id);
       return;
-    } else if (qt.taskType.equals(TaskType.PAYMENT) && !qt.order.paid.equals(PaymentStatus.TRYING)
-        || qt.taskType.equals(TaskType.MESSAGING) && (qt.messageType == 1
-        && !qt.order.messageSent.equals(MessageSent.NONE_SENT)
-        || qt.order.messageSent.equals(MessageSent.PAYMENT_FAIL)
-        || qt.order.messageSent.equals(MessageSent.PAYMENT_SUCCESSFUL))
-        || qt.taskType.equals(TaskType.EMPLOYEE_DB) && qt.order.addedToEmployeeHandle) {
+    } else if (this.isTaskDone(qt)) {
```
iluwatar, line 248: Maybe `isTaskDone` should be part of `QueueTask` interface?

iluwatar (changes requested):
Overall very nice refactorings. I had a couple of improvement ideas. If you could find time to address them then we can proceed with the merge.

PR #2539, opened 2023-06-19
MongoDB tests enabled using test containers. Resolves #2368

### LINKED ISSUE #2368: Enable the MongoDB tests in hexagonal pattern
Currently, the tests `MongoBankTest`, `MongoTicketRepositoryTest`, and `MongoEventLogTest` are disabled supposedly because they require a real MongoDB present. Let's try to re-enable the tests either by utilizing embedded MongoDB or using mocking techniques.

Acceptance criteria

- Hexagonal MongoDB tests are enabled and working

k1w1dev:
MongoDB tests now use test container which does require docker to be running.   This allows a mongodb container to start up before tests run. The host and port of the mongodb test container is injected into the system properties (existing pattern used by the tests) which is then used by the tests.

### DIFF
--- a/hexagonal/pom.xml
+++ b/hexagonal/pom.xml
@@ -51,6 +51,12 @@
       <groupId>org.mongodb</groupId>
       <artifactId>mongodb-driver-legacy</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.testcontainers</groupId>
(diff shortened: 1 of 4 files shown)


iluwatar:
This looks like a draft pull request. @k1w1dev please mark it ready when looking for review.

Issue #2507, opened 2023-04-20
Candidate for Known Uses of Proxy


VladimirCreator:
I propose the UIKit's [`UIAppearance`](https://developer.apple.com/documentation/uikit/uiappearance) protocol as the known use of the proxy design pattern.

thestbar:
Hey, I would like to work on this issue.

RobinKumar5986:
Hey I am new to the open source contribution can any one tell me how can I contribute and request and merge the pull requests...

damlaYasarr:
if no one solves it, can you assign it to me?

iluwatar:
@thestbar are you working on this?

RobinKumar5986:
Yes I am looking on this

thestbar:
@iluwatar I thought that I just had to add it in the known uses of this part, which I added here https://github.com/iluwatar/java-design-patterns/pull/2554. Is there anything else I should do? Of course if someone has started working on the extra stuff feel free to assign this to them!

SnehangshuChakraborty:
@iluwatar should this be closed ? looks like this is merged (asking this since this showed up under the good first issue label)

PR #2554, opened 2023-07-14
enhancement: added proxy pattern known use (#2507)

thestbar:
##Added New Known Use on Proxy Design Pattern

- As mentioned on issue #2507 Apple UIKit's UIAppearance use of proxy design pattern was added on README.md file under "Known uses" section. The [link ](https://developer.apple.com/documentation/uikit/uiappearance) that is used is the same that provided by the initiator of the issue.
- No extra changes were applied, since the fix is simple.


iluwatar (approved):

iluwatar:
@all-contributors please add @thestbar for docs

PR #2590, opened 2023-09-29
Enhanced pattern documentation

SkSadaf:
In this commit, I've improved the documentation for the following design patterns:
- Collection Pipeline: Added real-world examples, plain language explanation, and Wikipedia reference. Also included a programmatic example.
- Commander: Provided real-world scenarios, plain language description, and a Wikipedia link. Added a programmatic example where applicable.
- CQRS: Enhanced documentation with real-world illustrations, a simplified explanation in plain words, and a Wikipedia reference. Programmatic examples will be added as needed.
These updates align with the acceptance criteria, ensuring that each pattern's README.md file contains a comprehensive explanation.


iluwatar (changes requested):
The documentation looks otherwise good, but we can't accept Python code. The idea is to use the pattern example code in the explanation. See the other patterns for examples.

PR #2599, opened 2023-10-02
added printstream instead of foreach

JayDhurat:
Pull request title

- Clearly and concisely describes what it does
- Refer to the issue that it solves, if available

Pull request description


iluwatar (changes requested):
Parallel stream is not desired here. Often the overhead slows the code down and due to the parallel execution the prints may come out of order.

See https://www.baeldung.com/java-when-to-use-parallel-stream

PR #2709, opened 2023-10-18
Finished Explanation for Intercepting Filter #2236

JackH408:
Added intent, explanation, real world example, Wikipedia definition, programmatic example for single filter and multiple filters, applicability and consequences


iluwatar (approved):

iluwatar:

@all-contributors please add @JackH408 for docs

PR #2715, opened 2023-10-21
refactoring #1012: Format specifiers should be used instead of string concatenation.

kongleong86:
Format specifiers should be used instead of string concatenation.
Fix the issue reported by Sonar https://github.com/iluwatar/java-design-patterns/issues/1012 in this [link](https://sonarcloud.io/project/issues?resolved=false&rules=java%3AS3457&id=iluwatar_java-design-patterns&open=AW-DaC-fql1jQuQnKXrq).


special-case/src/main/java/com/iluwatar/specialcase/InvalidUser.java
```diff
@@ -42,6 +42,8 @@ public InvalidUser(String userName) {
 
   @Override
   public void show() {
-    LOGGER.info("Invalid user: " + userName);
+    if (LOGGER.isInfoEnabled()) {
+      LOGGER.info(String.format("Invalid user: %s", userName));
+    }
```
iluwatar, line 47: I would leave out the call `LOGGER.isInfoEnabled()`. The logger object already calls it internally and I think there is no reason to use it if we are not doing something very expensive inside the if-condition. Check this elsewhere within the changes as well.

iluwatar (changes requested):

special-case/src/main/java/com/iluwatar/specialcase/InvalidUser.java
kongleong86: Fixed.

iluwatar (approved):

PR #2725, opened 2023-10-23
improve tree node under iterator

DollyLiu5021:
improve iterator tree node

Pull request description

-add get height of tree
-add get count of tree node


iluwatar (approved):

iluwatar (changes requested):
Looks good, but Checkstyle checks are failing.

Issue #2257, opened 2022-10-29
Explanation for Resource Acquisition Is Initialization


iluwatar:
In this issue, let's add a proper explanation for the pattern.

Acceptance criteria

- The pattern's README.md has been amended with an explanation
- The explanation consists of
  - Real world example
  - In plain words
  - Wikipedia says
  - Programmatic example

ComplexOW:
Hi @iluwatar, I would like to contribute to this issue. Could you please assign this to me?

nooynayr:
I would like to contribute, would I be able to work on this? Also to confirm, is this in regards to java and not C++?

iluwatar:
Yes, everything in this repository is in Java

PR #2727, opened 2023-10-23
#2257 Added Explanation to Resource Acquisition is Initialisation README.md

nooynayr:
#2257 Added Explanation to Resource Acquisition is Initialisation README.md

Added explanations of the following:
- The explanation consists of
- Real world example
- In plain words
- Wikipedia says
- Programmatic example

### DIFF
--- a/resource-acquisition-is-initialization/README.md
+++ b/resource-acquisition-is-initialization/README.md
@@ -9,6 +9,117 @@ tag:
 ## Intent
 Resource Acquisition Is Initialization pattern can be used to implement exception safe resource management.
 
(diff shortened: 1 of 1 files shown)


iluwatar (changes requested):
The explanation should be written based on the code example. In the code, we have sliding door and treasure chest. The explanation now talks about car rental. Would it be possible to change it so that the explanation is in sync with the code?

iluwatar (approved):

iluwatar:

@all-contributors please add @nooynayr for docs

Issue #1300, opened 2020-07-07
Implicit Lock pattern


iluwatar:
## Description
The Implicit Lock design pattern ensures that only one thread can execute a particular section of code at a time without requiring explicit lock management by the developer. This is particularly useful in environments where multiple threads need to interact with shared resources, preventing data corruption and ensuring consistency. The pattern leverages synchronized methods or blocks to manage concurrent access automatically, simplifying the code and reducing the likelihood of concurrency-related bugs.

### Main Elements:
1. **Synchronization**: Use of synchronized methods or blocks to control access to critical sections of code.
2. **Automatic Lock Management**: The pattern abstracts lock management away from the developer, reducing the risk of errors.
3. **Concurrency Control**: Ensures that only one thread can execute the synchronized section at a time, preventing race conditions.

### References:
- [Martin Fowler's Catalog](https://martinfowler.com/eaaCatalog/implicitLock.html)
- [YouTube Explanation](https://www.youtube.com/watch?v=kDck7QuFzzg)
- [Project Contribution Guidelines](https://github.com/iluwatar/java-design-patterns/wiki)

### Acceptance Criteria:
1. Implement the Implicit Lock design pattern using synchronized methods or blocks in a new module.
2. Ensure thorough unit tests are created to validate the concurrency control and proper functioning of the pattern.
3. Update documentation to include an explanation of the Implicit Lock pattern, how it is implemented, and example use cases.

iluwatar:
Ok @djeong20

natashasrivastava:
Hey @iluwatar 
Can you assign this to me?

iluwatar:
Done @natashasrivastava, good luck

iluwatar:
It's not. Belongs to patterns of enterprise application architecture book.

Chant3ll3:
Hi @iluwatar! I want to work on this issue, could you assign it to me please?

itsamekadio:
Hi @iluwatar 
I would like to express my interest in working on the Implicit Lock Pattern issue. I believe this pattern will provide an excellent opportunity for me to work on thread synchronization and concurrency control in Java, and I’m excited to contribute to the project.

Please let me know if I can be assigned to this issue. Looking forward to your approval!

Best regards,
Abdelrahman Elkady

itsamekadio:
im working on this issue and have made some progress can u assign it to me please @iluwatar ?

ratech-dev:
Hey @iluwatar ,
[...]

PR #2730, opened 2023-10-24
feature: Add implicit lock pattern (#1300)

Chant3ll3:
Implicit Lock Pattern
- Implemented implicit lock pattern described in issue #1300

### DIFF
--- a/implicit-lock/README.md
+++ b/implicit-lock/README.md
@@ -1,0 +1,29 @@
+---  
+title: Implicit Lock
+category: Security
+language: en
+tag:
+ - Data access
+---  
+
+## Name / classification
+
+## Also known as
+
+## Intent
+
+## Explanation
+
+## Class diagram
+
+## Applicability
+
+## Tutorials
+
+## Known uses
+
+## Consequences
+
+## Related patterns
+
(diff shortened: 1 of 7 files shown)


implicit-lock/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+## Tutorials
+
+## Known uses
+
+## Consequences
+
+## Related patterns
+
+## Credits
```
iluwatar, line 29: The sections look good, but the body text is missing?

implicit-lock/src/main/java/com/iluwatar/implicitlock/App.java
```diff
@@ -0,0 +1,40 @@
...
+        // Start the threads
+        thread1.start();
+        thread2.start();
+
+        // Wait for threads to complete
+        try {
+            thread1.join();
+            thread2.join();
+        } catch (InterruptedException e) {
+            e.printStackTrace();
```
iluwatar, line 35: Use a proper logger (Lombok @Slf4j annotation)

iluwatar (changes requested):

Issue #2722, opened 2023-10-22
Translate Proxy pattern to portuguese


jppf22:
Translation of the Proxy pattern to Portuguese.
- Translation of the proxy pattern README to portuguese.
- This translation is part of the #2285 issue

jppf22:
Labels missing

jppf22:
Mistake on my part, though labels where assigned by author.

PR #2765, opened 2023-10-30
added missing diagram png for proxy pattern in PT translation

jppf22:
Added missing diagram png for proxy pattern in PT translation

- Added /etc/proxy.urm.png to the proxy Portuguese folder 

Pull request description
- The previous link to the png was broken since the png didn't exist in this folder.
- Simply created a png at the link's pointing location (like in the EN proxy pattern)


jppf22:
I forgot to add this png when solving issue #2722

Issue #2262, opened 2022-10-29
Explanation for Sharding


iluwatar:
In this issue, let's add a proper explanation for the pattern.

Acceptance criteria

- The pattern's README.md has been amended with an explanation
- The explanation consists of
  - Real world example
  - In plain words
  - Wikipedia says
  - Programmatic example

prasad-333:
@iluwatar can I try my hands on this one ?

PR #2792, opened 2024-01-01
Adds explanation for Sharding

prasad-333:
Issue https://github.com/iluwatar/java-design-patterns/issues/2262


iluwatar (changes requested):
Programmatic example missing from explanation

PR #2600, opened 2023-10-02
#2373 feature add: Vertical Slice Architecture.

sugan0tech:
Added the Vertical Slice Architecture #2373
 ̑

Issue #2373, opened 2022-12-02
Vertical slice architecture


iluwatar:
This is a big architectural pattern where components and dependencies are arranged based on vertical features/slices. The best descriptions are available at https://jimmybogard.com/vertical-slice-architecture/ and https://garywoodfine.com/implementing-vertical-slice-architecture/

Acceptance criteria

- Vertical slice architectural pattern implemented
- The implementation follows the guidelines in https://github.com/iluwatar/java-design-patterns/wiki/01.-How-to-contribute

thestbar:
I would like to start working on this new pattern! Could you please assign it to me?

sugan0tech:
@iluwatar  more over like this right? https://medium.com/sahibinden-technology/package-by-layer-vs-package-by-feature-7e89cde2ae3a

sugan0tech:
@thestbar  may I know the status of this epic?

iluwatar:
> @iluwatar more over like this right? https://medium.com/sahibinden-technology/package-by-layer-vs-package-by-feature-7e89cde2ae3a

Exactly. If @thestbar is not working on it, I'll assign to you.

thestbar:
> @thestbar  may I know the status of this epic?

Have done some little stuff but didn't have the time to finish it lately. @iluwatar feel free to assign it to @sugan0tech.

damlaYasarr:
if no one solves this, can you assign it to me?

sugan0tech:
@damlaYasarr I already made a PR, waiting for his review.

sugan0tech:
@iluwatar  I have made the PR,can you check it out?

damlaYasarr:
@sugan0tech I have a question.  I want to code facade design but I cant install mvn package. I created mvn project but which architecture that I have to choose?

sugan0tech:
> @sugan0tech I have a question.  I want to code facade design but I cant install mvn package. I mean , I created mvn project but which architecture that I have to choose?

You can architecture which is you think fits well with it. Also have some reference in other patterns and code changes in PR's as well.

sugan0tech:
Hi @iluwatar ,
I hope this message finds you well. I noticed that the PR I opened last year has become stale due to unforeseen circumstances on my end. I wanted to check if there's any ongoing progress or tracking related to this epic. If not, I would be happy to pick it up myself and see it through to completion.

PR #2828, opened 2024-03-10
 #2373 feature add: Vertical Slice Architecture.

sugan0tech:
Added the Vertical Slice Architecture https://github.com/iluwatar/java-design-patterns/issues/2373. previous PR: #2600


vertical-slice-architecture/README.md
```diff
@@ -0,0 +1,62 @@
+---
+title: Vertical Slice Architecture
+aka: Layer By Feature
```
iluwatar, line 3: Let's leave this line out since it's not in line with the pattern template we use

https://github.com/iluwatar/java-design-patterns/wiki/02.-Pattern-template

iluwatar: We can put the alternative name in its own chapter "Also known as"

vertical-slice-architecture/README.md
```diff
@@ -0,0 +1,62 @@
...
+In Plain Words
+
+> Vertical slice architecture is like organizing your toolbox.
+> Instead of having all your tools mixed together, you group them based on the type of task they perform.
+> This way, when you need a specific tool for a particular job,
+> you can quickly find it without rummaging through a jumble of items.
+
+> Similarly, in software development, vertical slice architecture involves organizing the codebase based on features. 
+> Each feature has its own self-contained set of components, making it easier to add, modify, or remove features without disrupting the entire application.
+
```
iluwatar, line 44: Would it be possible to add a minimal code example here?

iluwatar (changes requested):
Looks very good already. Added a couple of minor comments that we need to address before merging.

vertical-slice-architecture/README.md
sugan0tech: instead of code examples, can it be project file structure?

iluwatar: Yeah, I think we can improvise a bit here. Whatever describes the pattern most effectively.

iluwatar:
@sugan0tech some tests are failing and need to be fixed before merging

Issue #2532, opened 2023-05-30
ucls  file 


liulangjietou:
what generate ucls file software ?  l can't find it

iluwatar:
Those are ObjectAid UML Explorer files, but unfortunately the software seems to be discontinued. See https://stackoverflow.com/questions/71696763/how-to-install-objectaid-in-2022

PR #2830, opened 2024-03-15
refactor: Eliminate Optional from fields and parameters for async method invocation

### LINKED ISSUE #2829: Incorrect use of Optional class in Async Method Invocation
Optional class is used incorrectly in Async Method Invocation pattern.

The Optional class in Java is primarily intended to provide a type-level solution for representing optional values, instead of using null. It's a container object that may or may not contain a non-null value.  However, using Optional as a field in a class or as a parameter in a method is not recommended for several reasons:  

- Serialization: Optional is not serializable, so classes with Optional fields would not be serializable either. This can lead to problems if you're trying to send objects across a network or save them in a file for later use.  

- Increased complexity: Using Optional in these ways can make the code more complex and harder to understand. It's not immediately clear, for example, what it means for an Optional to be present or absent in the context of a field or method parameter.  

- Performance overhead: Optional creates an extra object for every wrapped value, which can lead to unnecessary memory usage and performance overhead.  

- Not intended use: The intended use of Optional is as a return type. The Java API designer, Brian Goetz, stated that Optional was designed to provide a limited mechanism for library method return types where there is a clear need to represent "no result", and using it as a field type is not recommended.  

Instead of using Optional as a field, it's often better to use a nullable field and add a comment that the field can be null. For method parameters, consider overloading the method or using a builder pattern if a parameter is optional.

Acceptance criteria

- Optional class fields and method parameters refactored

guqing:
### What this PR does?
Eliminate Optional from fields and parameters

This commit addresses the misuse of the Optional class within our Async Method Invocation pattern. Following #2829 , we have removed Optional from being used as class fields and method parameters.

**Key Changes:**
- Class fields previously typed as Optional have been converted to nullable types, with added documentation comments to indicate potential nullability.
- Method parameters that were Optional have been refactored.Removed `Optional<Exception>` from the `onComplete` method parameter of `AsyncCallback` and added an additional `onError(Exception)` method to call when an exception occurs

**Rationale:**
- Enhances serialization compatibility by removing non-serializable Optional fields.
- Reduces code complexity and improves readability by eliminating the ambiguity of Optional fields and parameters.
- Decreases performance overhead associated with the unnecessary object creation of Optional wrappers.

Fixes #2829

### DIFF
--- a/async-method-invocation/README.md
+++ b/async-method-invocation/README.md
@@ -57,7 +57,8 @@ public interface AsyncResult<T> {
 
 ```java
 public interface AsyncCallback<T> {
-  void onComplete(T value, Optional<Exception> ex);
+    void onComplete(T value);
+    void onError(Exception ex);
 }
 ```
 
--- a/async-method-invocation/etc/async-method-invocation.urm.puml
+++ b/async-method-invocation/etc/async-method-invocation.urm.puml
@@ -9,7 +9,8 @@ package com.iluwatar.async.method.invocation {
     + main(args : String[]) {static}
   }
   interface AsyncCallback<T> {
-    + onComplete(T, Optional<Exception>) {abstract}
+    + onComplete(T) {abstract}
+    + onError(Exception) {abstract}
   }
   interface AsyncExecutor {
     + endProcess(AsyncResult<T>) : T {abstract}
@@ -32,7 +33,7 @@ package com.iluwatar.async.method.invocation {
     ~ COMPLETED : int {static}
     ~ FAILED : int {static}
     ~ RUNNING : int {static}
-    ~ callback : Optional<AsyncCallback<T>>
+    ~ callback : AsyncCallback<T>
     ~ exception : Exception
     ~ lock : Object
     ~ state : int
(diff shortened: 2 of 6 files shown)


guqing:
Hi @iluwatar 
After extensive research, I have been unable to find suitable software to edit the file located at https://github.com/iluwatar/java-design-patterns/blob/master/async-method-invocation/etc/async-method-invocation.ucls. Following the issue in https://github.com/iluwatar/java-design-patterns/issues/2532, I attempted to download Eclipse and ObjectAid, yet these efforts were unsuccessful🫣. 
Could you please advise on how I should proceed?😅

iluwatar (approved):

iluwatar:
> Hi @iluwatar After extensive research, I have been unable to find suitable software to edit the file located at https://github.com/iluwatar/java-design-patterns/blob/master/async-method-invocation/etc/async-method-invocation.ucls and https://github.com/iluwatar/java-design-patterns/blob/master/async-method-invocation/etc/async-method-invocation.png . Following the issue in #2532, I attempted to download Eclipse and ObjectAid, yet these efforts were unsuccessful🫣. Could you please advise on how I should proceed?😅

Hey, don't worry about the .ucls-file. It is a duplicate since we primarily use .urm.puml files that are autogenerated. I'll remove the .ucls-file.

iluwatar:

@all-contributors please add @guqing for code

Issue #2880, opened 2024-03-31
Flaky test in Thread Local Storage


iluwatar:
The build randomly fails due to Thread-local Storage test failure. See the log below:

```
[INFO] -------------------------------------------------------
[INFO]  T E S T S
[INFO] -------------------------------------------------------
[INFO] Running ThreadLocalTest
Error:  Tests run: 2, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 6.086 s <<< FAILURE! -- in ThreadLocalTest
Error:  ThreadLocalTest.withoutThreadLocal -- Time elapsed: 3.052 s <<< FAILURE!
org.opentest4j.AssertionFailedError: expected: <false> but was: <true>
	at org.junit.jupiter.api.AssertionUtils.fail(AssertionUtils.java:55)
	at org.junit.jupiter.api.AssertFalse.assertFalse(AssertFalse.java:40)
	at org.junit.jupiter.api.AssertFalse.assertFalse(AssertFalse.java:35)
	at org.junit.jupiter.api.Assertions.assertFalse(Assertions.java:227)
	at ThreadLocalTest.withoutThreadLocal(ThreadLocalTest.java:68)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
	at org.junit.platform.commons.util.ReflectionUtils.invokeMethod(ReflectionUtils.java:725)
	at org.junit.jupiter.engine.execution.MethodInvocation.proceed(MethodInvocation.java:60)
	at org.junit.jupiter.engine.execution.InvocationInterceptorChain$ValidatingInvocation.proceed(InvocationInterceptorChain.java:131)
	at org.junit.jupiter.engine.extension.TimeoutExtension.intercept(TimeoutExtension.java:149)
	at org.junit.jupiter.engine.extension.TimeoutExtension.interceptTestableMethod(TimeoutExtension.java:140)
	at org.junit.jupiter.engine.extension.TimeoutExtension.interceptTestMethod(TimeoutExtension.java:84)
[...]

PR #2884, opened 2024-04-01
Thread local storage test change

junhkang:
Flaky test in Thread Local Storage #2880

Create independent threadLocal to make values unsharable and unchangable by other threads.
(Not sure this change is what I supposed to change, let me know if there is something that I have to consider or change :))


iluwatar (approved):

iluwatar:
Looks good! Thank you for looking into this 🎉

Issue #2968, opened 2024-05-21
Version number pattern update overwritten


MapleChenf:
![image](https://github.com/iluwatar/java-design-patterns/assets/48958773/8ce6a191-dbf3-45cf-a868-7bb6a65b8a87)
If two clients judge by version number at the same time, will it cause write overwrite。

Avinash2110:
Hi! I would like to work on this issue.

PR #2980, opened 2024-06-02
fix: Version number pattern update overwritten #2968

Avinash2110:

Currently the version number pattern is not thread safe in case multiple updates are done at the same time. To resolve this I have used ConcurrentHashMap and written the critical code inside the synchronized block with a lock to avoid dirty read and  overwriting.


iluwatar (approved):

iluwatar:

@all-contributors please add @Avinash2110 for code

PR #2988, opened 2024-06-05
Fix : Issue#2976 - Refactoring for the code smells.

ankurnotwarikoo:
Issue#2976 Refactor Layered Architecture.
Along with the smells mentioned in the issue, there were some another smells identified which was fixed.
For eg. CodeBakingServiceImpl


iluwatar (changes requested):
Check that the changes are also reflected to `README.md` programmatic example

PR #2995, opened 2024-06-11
Update README.md

azarboon:

Before diving into any of the patterns, readers should be reminded of two fundamental laws in software architecture:

1. Everything is a trade-ff
2. "Why is more important than the how"

So, readers face the nuances and reality of these patterns from the beginning. These two laws are coined by two thought leaders in software architecture: Mark Richards and Neal Ford. They have explained these two laws in various conference talks and books. For example, Here you can read about these two laws here:

https://www.infoq.com/podcasts/software-architecture-hard-parts/

Also, here is a book for reference:
https://a.co/d/fKOodW9


iluwatar:
Thanks, but I like the text as it is 🙂

Issue #2977, opened 2024-06-01
Fix busy-waiting loops


iluwatar:
### Description

Busy-waiting, or spinning, is a technique where a process repeatedly checks to see if a condition is true, such as whether keyboard input or a lock is available. While it might seem like a good idea, it has several significant drawbacks:  

1. CPU Usage: Busy-waiting can consume a lot of CPU time. While a process is busy-waiting, it keeps the CPU busy. This can prevent other processes from running and can lead to high CPU usage, especially if the condition being checked doesn't become true for a long time.  
2. Performance: Busy-waiting can lead to performance issues. If a process is busy-waiting, it's not doing any useful work. This can slow down the overall performance of your application.  
3. Responsiveness: Busy-waiting can make your application less responsive. If a process is busy-waiting, it might not be able to respond to user input or other events in a timely manner. 
4. Power Consumption: Busy-waiting can lead to increased power consumption, especially on battery-powered devices. This is because the CPU is kept busy and is not allowed to enter a low-power state.  

Instead of busy-waiting, it's generally better to use some form of event-driven programming or blocking. This allows your process to sleep until the condition it's waiting for becomes true, which can save CPU time, improve performance, and make your application more responsive.

Busy-waiting loops are at least in these locations:

1. Server Session / App.java
2. Twin / BallThread.java
3. Log Aggregation / LogAggregator.java
4. Commander / Retry.java
5. Retry / Retry.java
6. Retry / RetryExponentialBackoff.java
7. Queue-Based Load Leveling / ServiceExecutor.java

### Acceptance Criteria

- Busy-waiting loops refactored
- README.md of the affected patterns revised as needed

Mainak2002:
can this issue be assigned to me?

TATIKONDAVENKATESH:
i am new here, i want to contribute to this project

Mehdi-17:
Hello, i'm available to work on it

Mehdi-17:
Thanks, I've just started working on this.

shalini-bhandari:
Hello, this is my first time in open source. 
Is this issue still open? and how to go about it?

AdityaBhendavadekar:
hey @iluwatar i want to work on this. please assign this issue to me

keshavMM004:
[...]

PR #2999, opened 2024-06-18
Fix busy-waiting loops #2977

shree243:


iluwatar (changes requested):
The build is failing due to Checkstyle errors

PR #3001, opened 2024-06-24
fix: Refactored Visitor #2975 

### LINKED ISSUE #2975: Refactor Visitor
### Description

`VisitorTest` has some code smells regarding `Optional` usage.

1. `Optional<String>` has been used as class field type.
2. `Optional<String>` has been used as parameter type.

`Optional` is primarily intended for use as a method return type where there is a clear need to represent "no result," and where using `null` is likely to cause errors. A variable whose type is `Optional` should never itself be null; it should always point to an `Optional` instance.

### Acceptance Criteria

- Code smells refactored

Joel-Dandin:

This PR refactors the VisitorTest class to address code smells related to the misuse of Optional<String> for class fields and constructor parameters. Optional is intended for method return types to indicate "no result," and not for class fields or parameters.

Close #2975


iluwatar (approved):

iluwatar:

@all-contributors please add @Joel-Dandin for code

PR #3016, opened 2024-07-19
Fix busy-waiting loops for Server Session / App.java

AndyC5H10O5:

In the original code, busy waiting is implemented through an infinite loop and Thread.sleep. This method has the following issues:
1. Resource waste: Even without expired sessions, threads will continue to run, consuming CPU resources.
2. Poor scalability: As the number of sessions increases, this method consumes more CPU resources and reduces system performance.
In the optimized code, busy waiting is replaced by SchechedExecutorService, which executes tasks periodically through event driven methods.

### DIFF
--- a/server-session/src/main/java/com/iluwatar/sessionserver/App.java
+++ b/server-session/src/main/java/com/iluwatar/sessionserver/App.java
@@ -31,6 +31,9 @@ import java.time.Instant;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.Map;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
 import lombok.extern.slf4j.Slf4j;
 
 /**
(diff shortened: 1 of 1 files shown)


server-session/src/main/java/com/iluwatar/sessionserver/App.java
```diff
@@ -75,37 +78,34 @@ public static void main(String[] args) throws IOException {
...
-                  iterator.remove();
-                }
+  private static void startSessionExpirationTask() {
+    ScheduledExecutorService scheduler = Executors.newScheduledThreadPool(1);
+    Runnable sessionExpirationChecker = () -> {
+      try {
+        LOGGER.info("Session expiration checker started...");
+        Instant currentTime = Instant.now();
+        synchronized (sessions) {
+          synchronized (sessionCreationTimes) {
```
iluwatar, line 93: Could we use ConcurrentHashMap for sessions and sessionCreationTimes allowing us to remove the need for synchronization blocks?

server-session/src/main/java/com/iluwatar/sessionserver/App.java
```diff
@@ -75,37 +78,34 @@ public static void main(String[] args) throws IOException {
...
+            Iterator<Map.Entry<String, Instant>> iterator = sessionCreationTimes.entrySet().iterator();
+            while (iterator.hasNext()) {
+              Map.Entry<String, Instant> entry = iterator.next();
+              if (entry.getValue().plusMillis(SESSION_EXPIRATION_TIME).isBefore(currentTime)) {
+                sessions.remove(entry.getKey());
+                iterator.remove();
```
iluwatar, line 99: Add debug logging for session removal

server-session/src/main/java/com/iluwatar/sessionserver/App.java
```diff
@@ -75,37 +78,34 @@ public static void main(String[] args) throws IOException {
...
-          LOGGER.error("An error occurred: ", e);
-          Thread.currentThread().interrupt();
         }
+        LOGGER.info("Session expiration checker finished!");
+      } catch (Exception e) {
+        LOGGER.error("An error occurred: ", e);
       }
-    }).start();
+    };
+    scheduler.scheduleAtFixedRate(sessionExpirationChecker, 0, SESSION_EXPIRATION_TIME, TimeUnit.MILLISECONDS);
```
iluwatar, line 109: We should also shut down the scheduler when the program exits. Something like this:

```
Runtime.getRuntime().addShutdownHook(new Thread(() -> {
        scheduler.shutdown();
        try {
            if (!scheduler.awaitTermination(60, TimeUnit.SECONDS)) {
                scheduler.shutdownNow();
            }
        } catch (InterruptedException e) {
            scheduler.shutdownNow();
        }
    }));
```

iluwatar (changes requested):

PR #3041, opened 2024-10-05
Update NullNode.java

Dibyajyoti1515:
After merge this the getname() function return "There is no node" not give null value


Ritabrata1080 (approved):
LGTM

PR #3043, opened 2024-10-05
Create App.java

Dibyajyoti1515:
Chain of Responsibility Pattern in Java: Building Robust Request Handling Mechanisms

### DIFF
--- a/localization/zh/chain/App.java
+++ b/localization/zh/chain/App.java
@@ -1,0 +1,93 @@
+@Getter
+public class Request {
+
+    private final RequestType requestType;
+    private final String requestDescription;
+    private boolean handled;
+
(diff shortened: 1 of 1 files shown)


localization/zh/chain/App.java
```diff
@@ -0,0 +1,93 @@
+@Getter
+public class Request {
+
+    private final RequestType requestType;
+    private final String requestDescription;
+    private boolean handled;
+
+    public Request(final RequestType requestType, final String requestDescription) {
+        this.requestType = Objects.requireNonNull(requestType);
```
Ritabrata1080, line 9: Instead of Objects.requireNonNull(), can we use @NonNull annotation? Please check once the implementation.

PR #3101, opened 2024-11-30
Fixed busy waiting in points 1,2,3,7

alhusseain:
Close issue #2977
Implemented wait and notify functionalities on threads that have busy waiting issues

No. 1 : Server Session / App.java

sessionExpirationTask now waits until at least a single login request is sent, since there would be unnecessary overhead if it runs with no login requests available.

No. 2 : Twin / BallThread.java

run() now waits if suspendMe is called, and resumes if resumeMe is called, since there would be unnecessary overhead if run() is running when ball item is suspended. ( no need to always check if suspended)

No. 3: Log Aggregation / LogAggregator.java
startBufferFlusher runs if there is at least one log is present, similiar to No.1's fix

No. 7: Queue-Based Load Leveling / ServiceExecutor.java

similiar to No.3 and No.1, at least a single msg is added to queue for it to work, until then run() waits.

I did not refactor No. 4,5,6 since I think no unnecessary overhead is present ( as the retry pattern implemented loops until certain number of errors occur, doesn't loop needlessly)

I would much appreciate your feedback, which I would consider and act upon quickly.


alhusseain:
Quality Gate Gives a Reliability Rate D to a file I did not Change or commit (CandyGame.java) . Also I did not change any thing in LogoutHandlerTest.java So I do not know where the conflict is.

twin/src/main/java/com/iluwatar/twin/BallThread.java
```diff
@@ -42,37 +42,58 @@ public class BallThread extends Thread {
...
+    try {
+      while (isRunning) {
+        if (isSuspended) {
+          synchronized (lock) {
+            lock.wait();
+          }
+        } else {
+          twin.draw();
+          twin.move();
+          Thread.sleep(250);
```
iluwatar, line 60: This `Thread.sleep(250)` is the main problem. Instead of using a thread, we could utilize the same aforementioned `ScheduledExecutorService `

iluwatar (changes requested):

twin/src/main/java/com/iluwatar/twin/BallThread.java
alhusseain: On it, Thanks for the feedback

alhusseain: I have implemented all requested changes except only in BallThread.java. in BallThread.java, There is no need to run it every fixed amount of time. Instead, waking the thread up only when there is a request to resume and making it wait when there is a request to suspend is sufficient to reduce overhead. This way, I implement event-based logic.

iluwatar:
I'm sorry, but it doesn't look correct. The solution needs a complete rewrite, so I'll close this.

PR #3114, opened 2024-12-06
Identity Field Pattern #1298

### LINKED ISSUE #1298: Identity Field pattern
https://www.sourcecodeexamples.net/2018/05/identity-field-pattern.html

Rehab-M-Esmail:

This pattern solves the problem with the primary key issue which generates the Id for each instance that extends it. It is essential when the uniqueness of certain attributes is required to maintain data integrity, avoid conflicts, and support efficient lookups.
Close #1298


Rehab-M-Esmail:
@iluwatar I will modify them

iluwatar:
The build is failing. Please ensure that all the checks pass before requesting review.

Issue #1312, opened 2020-07-07
Row Data Gateway pattern


iluwatar:
https://www.sourcecodeexamples.net/2018/04/row-data-gateway.html

tao-ny:
Could I work on this? @iluwatar

ohbus:
Sure thing @tao-ny 

Please mention a timeline for when can we expect a PR against this issue.

tao-ny:
I will finish it within a week.

ShawnMaxShen:
Hi, Can I work on this issue?

iluwatar:
Ok @ShawnMaxShen

BorisBDing:
hi, anything I can work for this issue?

ShawnMaxShen:
@BorisBDing  bro, we can have a discussion privately. Your comment may make maintainer a little confusing. 
(For maintianer, me and Boris are working together~)

BorisBDing:
Hi, Can I work on this issue?

nour-awad:
@iluwatar can you assign me to this task?

PR #3120, opened 2024-12-07
Modified the row data gateway design pattern (Issue #1312)

nour-awad:
### Issue #1312 
**This PR addresses issues encountered while running tests, particularly related to database interactions and data retrieval. However, the issue is not completely resolved, as I faced problems during the `mvn clean install` process and encountered database-related errors. These challenges are still being worked on, and further troubleshooting is required to ensure full resolution.**

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -201,6 +201,7 @@
     <module>currying</module>
     <module>serialized-entity</module>
     <module>identity-map</module>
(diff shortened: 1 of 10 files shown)


nour-awad:
Screenshot 2024-12-07 at 15 17 32
If you can help me solve this I will be so grateful.

PR #3126, opened 2024-12-07
fixing-busy waiting loop #server session

Nothere998:
What problem does this PR solve?

This PR addresses the issue of busy-waiting in the session expiration task of the Server Session Pattern. Previously, the session expiration thread continuously checked for expired sessions without any pauses, which led to unnecessary CPU consumption.

Key improvements:
Replaced the busy-waiting loop with a Thread.sleep() mechanism, ensuring the thread pauses and consumes no CPU resources between checks.
Added graceful handling of thread interruptions for proper thread lifecycle management.

This change significantly reduces the overhead caused by constantly checking for session expiration, making the session management more efficient and resource-friendly.

### DIFF
--- a/server-session/README.md
+++ b/server-session/README.md
@@ -64,29 +64,30 @@ public class App {
 
   private static void sessionExpirationTask() {
     new Thread(() -> {
-      while (true) {
-        try {
-          Thread.sleep(SESSION_EXPIRATION_TIME);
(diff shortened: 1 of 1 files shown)


iluwatar:
Changing the README.md file doesn't fix the code because the README.md file is a documentation file, not the code itself.

PR #3127, opened 2024-12-07
Fix busy-waiting loops-#Log Aggregation 

Nothere998:
in microservices, logs are often collected using busy waiting loops, where the system constantly checks for new logs. This leads to several issues:

Wasting Resources: The system keeps checking for new logs, using up CPU resources even when there are no logs to process.
Scalability Problems: As the number of services grows, busy waiting loops become inefficient and slow, causing delays and performance issues.
Increased Latency: Logs are delayed because the system is busy checking for new data rather than processing logs as they arrive.

This PR solves these problems by switching to a more efficient event-driven log aggregation system. Instead of constantly checking for new logs, the system reacts only when new logs are available, reducing unnecessary CPU usage and improving scalability and performance

### DIFF
--- a/microservices-log-aggregation/README.md
+++ b/microservices-log-aggregation/README.md
@@ -48,37 +48,85 @@ The `CentralLogStore` is responsible for storing the logs collected from various
 
 ```java
 public class CentralLogStore {
+    private final List<LogEntry> logs = new ArrayList<>();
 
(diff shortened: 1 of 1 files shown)


iluwatar (changes requested):
Unfortunately the changes seem to affect only README.md files, so the software behavior is still unfixed

PR #3129, opened 2024-12-07
fixing-twin

Nothere998:
Solution: used Thread Synchronization with wait() and notify()
By employing thread synchronization mechanisms (wait and notify), the BallThread can be made to pause and resume efficiently without busy waiting.


iluwatar:
Not sure what problem this solves.

Additionally, the build is failing. Please ensure that all the checks pass before requesting review.

Issue #76, opened 2015-06-23
Scheduler pattern


iluwatar:
https://en.wikipedia.org/wiki/Scheduler_pattern

mikulucky:
Can I take this one please.

iluwatar:
Please, go ahead @mikulucky :+1:

iluwatar:
@mikulucky are you still working on this?

iluwatar:
This issue is free for taking again.

mpodlodowski:
I'd like to take this one.

iluwatar:
@mpodlodowski great, you're good to go!

iluwatar:
@mpodlodowski  are you still implementing this?

iluwatar:
This issue is free again

tiennm99:
Can I take it @iluwatar?

iluwatar:
Updated task description

ipaulaandreea:
@iluwatar Can you assign me on this task, please?

MME517:
Can you assign me to this issue, please?

Nothere998:
Hello @iluwatar,i would love to contribute to this issue.could you please assign me to it?

iluwatar:
I think @MME517 has this under construction

PR #3135, opened 2024-12-07
Scheduler pattern #76

Nothere998:
In many applications, it is necessary to manage and execute tasks (or jobs) at specific intervals or under certain conditions. Examples include running periodic background tasks, sending scheduled notifications, or automating repetitive processes. However, managing these tasks manually can lead to several challenges:

Lack of Modularity: Without a well-structured design, the logic for task execution may become entangled with application logic, reducing code readability and maintainability.
Hardcoding of Triggers: Defining triggers inline or directly in the task logic makes it hard to reuse or update the triggers without modifying the entire task.
No Centralized Management: The absence of a centralized scheduler results in scattered execution logic, making debugging and scalability difficult.
Inflexibility: Hardcoded schedules or conditions prevent the easy addition or modification of tasks and triggers.

The goal was to design a scheduling pattern that decouples jobs from their execution triggers, promotes reusability, and centralizes the scheduling logic.


schedule pattern
```diff
@@ -0,0 +1,69 @@
...
+    private List<Trigger> triggers;  
+    public Scheduler() {  
+        this.jobs = new ArrayList<>();  
+        this.triggers = new ArrayList<>();  
+    }  
+    public void addJob(Job job, Trigger trigger) {  
+        jobs.add(job);  
+        triggers.add(trigger);  
+    }  
+    public void start() {
```
akashpawar6474, line 20: actually we use spring boot and the framework itself provides centralized system and the dev just needs to use specific stereotype annotations to schedule job with configs

PR #3154, opened 2024-12-12
Update pom.xml

yuvashrikarunakaran:
l Added the license-maven-plugin with the check goal to validate licenses in source files


iluwatar:
The build is failing. Please ensure that all the checks pass before requesting review.

PR #3155, opened 2024-12-12
Update pom.xml

yuvashrikarunakaran:
Artifact Name: Changed from anti-corruption-layer to corruption-layer to reflect the updated context


iluwatar:
And what problem does this solve?

PR #3156, opened 2024-12-12
Update docker-compose.yml

yuvashrikarunakaran:
Named Volumes are managed by Docker and optimized for performance. Docker can manage the lifecycle and cleanup of volumes, unlike host-mounted directories that might require manual cleanup.


iluwatar:
What problem does this PR solve?

PR #3274, opened 2025-05-03
Refactor Task execution to support asynchronous callbacks

HY-love-sleep:
### Summary

This PR refactors the existing Task execution model to support asynchronous callbacks. The `executeWith` method has been modified to execute tasks asynchronously using `CompletableFuture`, allowing the main thread to remain non-blocking. Upon task completion, the provided callback is invoked.

### Changes

- Replaced synchronous task execution with asynchronous execution using `CompletableFuture.runAsync()`.
- Updated method signatures and logic to accommodate asynchronous behavior.
- Enhanced code comments to clarify the asynchronous execution flow.

### Motivation

The original implementation was synchronous, potentially blocking the main thread. By introducing asynchronous execution, we improve the responsiveness and scalability of the application, especially in scenarios involving long-running tasks.In addition, Microsoft's Event-based Asynchronous Pattern (EAP) also emphasizes the use of callbacks to handle notifications and result processing after asynchronous operations.

### Testing

- Unit tests have been updated to verify asynchronous execution and callback invocation.
- Performance tests indicate improved responsiveness with non-blocking behavior.

Please review the changes and provide feedback.


callback/src/main/java/com/iluwatar/callback/Task.java
```diff
@@ -24,15 +24,20 @@
...
   final void executeWith(Callback callback) {
-    execute();
-    Optional.ofNullable(callback).ifPresent(Callback::call);
+    CompletableFuture.runAsync(
+        () -> {
+          execute();
+          if (callback != null) {
+            callback.call();
+          }
+        });
```
iluwatar, line 40: In general, looks good! However, I'd like to keep the synchronous callbacks (previous functionality) as well.

Could we demonstrate both sychronous and asynchronous callback in the example? The guidance would be something like

- If your operation is short and straightforward, keep it synchronous.
- If the operation can block or is resource-intensive, prefer asynchronous callbacks to maintain application responsiveness.

iluwatar (changes requested):

HY-love-sleep:
Hi @iluwatar , thanks for your feedback!

I’ve just pushed updates to:

- Keep both synchronous (executeWith) and asynchronous (executeAsyncWith) callbacks.

- Replace Thread.sleep with CompletableFuture.join()/orTimeout() for async examples.

- Refactor tests to use AtomicInteger and CompletableFuture for stable, non-blocking waits.


Please let me know if there’s anything else you’d like me to adjust!

Thanks again for reviewing.

iluwatar:
@all-contributors please add @HY-love-sleep for code

PR #3300, opened 2025-07-08
feat: add factory enum

huunghia98er:
# Pull Request Template

## What does this PR do?

1. Refactors the existing Factory Pattern implementation to use Factory Enum Pattern.

- Moves object creation logic into enum (FileProcessorType)

- Removes the need for a separate factory class

2. Applies singleton via InstanceHolder for each processor

## Why is this change necessary?

- Simplifies the codebase by eliminating boilerplate factory classes

- Improves readability and maintainability

- Leverages type-safe enum-based instantiation

- Aligns with modern Java practices (e.g. Enum Singleton)


factory-enum/src/test/java/com/iluwatar/factory/FileProcessorTest.java
```diff
@@ -0,0 +1,56 @@
...
+
+import static org.junit.jupiter.api.Assertions.*;
+
+import org.junit.jupiter.api.Test;
+
+class FileProcessorTest {
+
+  @Test
+  void shouldReturnGoldCoinInstance() {
+    final var goldCoin = FileProcessorType.PDF.getInstance();
```
ronodhirSoumik, line 35: just to know - any particular reason to use final here for local variable?
and same for the other test cases

factory-enum/src/main/java/com/iluwatar/factory/ExcelProcessor.java
```diff
@@ -0,0 +1,45 @@
...
+package com.iluwatar.factory;
+
+import lombok.AccessLevel;
+import lombok.NoArgsConstructor;
+
+/** ExcelProcessor implementation. */
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+public class ExcelProcessor implements FileProcessor {
+
+  static final String DESCRIPTION = "This is a excel processor.";
```
ronodhirSoumik, line 34: minor suggestion-
this better be private and class property should be accesses only through `getDescription`

Issue #3225, opened 2025-03-30
Implement Thread-Specific Storage Pattern


iluwatar:
## Description

Thread-Specific Storage is a concurrency design pattern where each thread retains its own instance of a shared object, typically achieved using `ThreadLocal<T>` in Java. By isolating data to each thread, you avoid synchronization overhead and minimize concurrency issues. Common use cases include storing thread-specific contexts, caching stateful objects like `DateFormat`, or managing per-thread counters without risking data corruption or race conditions.

### Key Elements

- **Isolation of State:** Each thread has its own copy of the data, reducing shared mutable state.
- **ThreadLocal Utility:** Java’s `ThreadLocal<T>` class provides a straightforward way to store data private to each thread.
- **Initialization & Cleanup:** Properly initializing and cleaning up thread-local data is crucial to prevent memory leaks.
- **Practical Use Cases:** Storing per-thread data such as locale-specific formatters, current transaction context, or local caches.

## References

1. *Pattern-Oriented Software Architecture, Volume 2: Patterns for Concurrent and Networked Objects*
2. [Java Design Patterns - Contribution Guidelines](https://github.com/iluwatar/java-design-patterns/wiki)

## Acceptance Criteria

- [ ] A new module or package named `thread-specific-storage` (or similar) is added.
- [ ] Demonstrates a clear use case of storing and retrieving thread-local data (e.g., per-thread state).
- [ ] Includes a readme (or `.md` file) explaining how the pattern works, with code examples or diagrams.
- [ ] Adheres to the repository’s coding style and naming conventions.
- [ ] Includes unit tests verifying behavior in a multi-threaded scenario.
- [ ] Passes all CI checks and meets the project’s contribution guidelines.

CMD137:
Hello @iluwatar .I would like to work on this .Could you assign it to me? It would be a great opportunity to learn!

RitikaMathur19:
Is someone working on this already , No one is assigned yet. Else please assign it to me.

kdt523:
Hi, I’d like to work on this issue . Could you please assign it to me?

iluwatar:
https://github.com/iluwatar/java-design-patterns/pull/3422

PR #3384, opened 2025-10-22
thread-specific-storage

quantambites:
# Pull Request Template

## What does this PR do?

Thread-Specific Storage is a concurrency design pattern where each thread retains its own instance of a shared object, typically achieved using ThreadLocal<T> in Java. By isolating data to each thread, you avoid synchronization overhead and minimize concurrency issues. Common use cases include storing thread-specific contexts, caching stateful objects like DateFormat, or managing per-thread counters without risking data corruption or race conditions.

Key Elements
Isolation of State: Each thread has its own copy of the data, reducing shared mutable state.
ThreadLocal Utility: Java’s ThreadLocal<T> class provides a straightforward way to store data private to each thread.
Initialization & Cleanup: Properly initializing and cleaning up thread-local data is crucial to prevent memory leaks.
Practical Use Cases: Storing per-thread data such as locale-specific formatters, current transaction context, or local caches.

#3225


iluwatar:
Already implemented in https://github.com/iluwatar/java-design-patterns/pull/3422

Issue #2686, opened 2023-10-15
Microservice pattern: Server-side service discovery


iluwatar:
https://microservices.io/patterns/server-side-discovery.html

lokytech5:
can you assign me this

lokytech5:
hello, please specify if i can use spring boot to implement the design or core java.

iluwatar:
Yes, Spring Boot can be used. It has been used in some other patterns as well.

lokytech5:
currently on it

ankurnotwarikoo:
How do you plan to implement this ? 
What all microservices do you want to discover can I get a gist of those ?

iluwatar:
The issue is currently unassigned. We are open for proposals 😀

AmanH7618:
@iluwatar Can you please assign this to me?

RitikaMathur19:
Am working on this will share the PR soon. @iluwatar please assign it to me

PR #3402, opened 2025-11-20
feature-server-side-service-discovery #2686

RitikaMathur19:
Microservice pattern: Server-side service discovery #2686

## What does this PR do?

This is a feature for Server-side service discovery pattern
The Server-Side Service Discovery pattern is a microservice architecture pattern that provides a centralized mechanism for services to register themselves and for consumers to discover available services dynamically. This pattern enhances system scalability and flexibility by decoupling service consumers from the physical locations of service providers.

#2686

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -215,6 +215,7 @@
     <module>service-locator</module>
     <module>service-stub</module>
     <module>service-to-worker</module>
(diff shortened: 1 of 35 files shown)


iluwatar (changes requested):
The build is failing

PR #3410, opened 2025-12-05
test: add assertion for executing App.main with empty args

AlMartinezUy:
# Pull Request Template

## What does this PR do?

This PR adds an additional assertion to the `AppTest` test class in the
`abstract-document` module. The new assertion verifies that calling
`App.main` with an empty `String[]` argument does not throw any exception.

This complements the existing assertion that checks the behavior when the
`main` method is invoked with a `null` argument.

No production code is modified. This PR solely improves test coverage and
validates expected behavior under different argument conditions.


AlMartinezUy:
Hi! I've updated the branch with the latest master changes. One of the checks is now passing successfully. The Java PR Builder failure seems unrelated to my changes, as this PR only adds a small test case for empty arguments in the abstract-document module. Could a maintainer please take a look when possible? Thanks!

PR #3449, opened 2026-03-20
docs: clarify Immutable Object pattern in Value Object README (fixes #3448)

### LINKED ISSUE #3448: Missing Pattern Immutable
Currently there is no pattern treating immutable classes. 

Is that a mistake or is there a reason not to consider this as a design pattern? In the latter case I would not expect "Private Class Data" to be mentioned.

Senrian:
## Fixes Issue #3448: Missing Pattern Immutable

The issue requests adding the Immutable Object pattern to the repository. This PR addresses it by:

1. **Promoting "Immutable Object" to the first entry in "Also known as"** - making it clear that Value Object and Immutable Object are the same pattern
2. **Updating the Intent section** - Renamed to "Intent of Value Object / Immutable Object Design Pattern" and added explicit mention of the Immutable Object pattern benefits

Note: The Value Object pattern already implements the Immutable Object pattern (using Lombok `@Value` annotation). This PR makes that relationship explicit in the documentation.

### DIFF
--- a/page-object/src/main/java/com/iluwatar/pageobject/App.java
+++ b/page-object/src/main/java/com/iluwatar/pageobject/App.java
@@ -77,7 +77,8 @@ public final class App {
       } else {
         // Java Desktop not supported - above unlikely to work for Windows so try the
         // following instead...
-        Runtime.getRuntime().exec("cmd.exe start " + applicationFile);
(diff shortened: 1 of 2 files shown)


sieberss:
In my understanding Value Object and Immutable Object are different patterns.
Value Object is implemented by using immutable objects, it does not provide a solution for constructing them.

PR #3509, opened 2026-06-04
feat: implement Immutable pattern (Fixes #3448)

### LINKED ISSUE #3448: Missing Pattern Immutable

alex052525:
## What does this PR do?
  
  Adds the **Immutable** design pattern module, which was missing from the repository.
  
  - Fixes #3448
  
  ### Implementation highlights
  
  - `ImmutableUser` — `final class` with `final` fields, defensive copy via `List.copyOf()`,
    and wither methods (`withName`, `withAge`, `withRoles`) that return new instances instead of mutating state
  - `App` — runnable demo showing immutability and defensive copy in action 
  - `ImmutableUserTest` — 10 JUnit 5 tests covering construction, wither behaviour,
    original-object invariance, defensive copy, and unmodifiable getter
  - `etc/immutable.urm.puml` — PlantUML class diagram
  - `README.md` — full pattern documentation following project conventions

### DIFF
--- a/immutable/README.md
+++ b/immutable/README.md
@@ -1,0 +1,133 @@
+---
+title: "Immutable Pattern in Java: Building Thread-Safe Objects"
+shortTitle: Immutable
+description: "Learn the Immutable pattern in Java with real-world examples, class diagrams, and tutorials. Understand how to create objects that cannot be modified after construction."
+category: Idiom
+language: en
+tag:
+  - Immutability
+  - Thread safety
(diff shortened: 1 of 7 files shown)


iluwatar:
@all-contributors please add @alex052525 for code

PR #3543, opened 2026-07-07
Add Backends For Frontends (BFF) pattern (#300)

### LINKED ISSUE #300: Backends for Frontends pattern
**Description:**
The Backends for Frontends (BFF) design pattern is intended to create separate backend services for different user interfaces or clients. This pattern is particularly useful when different clients (such as mobile apps, desktop applications, and web applications) have distinct needs and require tailored backend interactions. Implementing BFF helps to ensure that each client gets exactly the data it requires in the optimal format, improving performance and maintainability.

**Main Elements of the Pattern:**
1. **Client-Specific Backend:** Each frontend has a dedicated backend service that caters to its specific needs.
2. **Data Aggregation:** The BFF aggregates data from various sources and formats it appropriately for the frontend.
3. **Decoupling:** This pattern decouples the frontend from the complexities of backend systems, providing a clean API tailored to the frontend's requirements.
4. **Optimized API:** Each backend service provides an optimized API for its respective client, improving performance and user experience.

**References:**
- [Backends for Frontends (BFF) Pattern](https://samnewman.io/patterns/architectural/bff/)
- [Microservices Patterns: With examples in Java](https://www.amazon.com/Microservices-Patterns-examples-Chris-Richardson/dp/1617294543)
- [Github Contribution Guidelines](https://github.com/iluwatar/java-design-patterns/wiki)

**Acceptance Criteria:**
1. **Client-Specific Backend Services:** Implement distinct backend services for at least two different clients (e.g., web and mobile).
2. **Data Aggregation:** Ensure that the BFFs aggregate data from multiple services and format it specifically for their respective clients.
3. **Optimized APIs:** Create clean and optimized APIs for each backend service to improve the performance and user experience of the respective frontends.

AnveshSrivastava:
## Summary

Closes #300

Implements the Backends For Frontends (BFF) pattern as a new module,
`backends-for-frontends`. Two client-specific gateways — `MobileBff` and
`DesktopBff` — sit in front of a shared set of downstream microservices
(`AuthService`, `CartService`, `OrderService`, `SupplierService`) and each
aggregates only the services its own client needs, reshaping the result into
a response tailored to that client:

- `MobileBff` calls Auth + Cart + Order → returns a lean `MobileDashboardResponse`
  suited to a phone screen.
- `DesktopBff` calls Auth + Order + Supplier → returns a richer
  `DesktopDashboardResponse` with back-office detail a mobile client never needs.

This satisfies the issue's acceptance criteria: distinct client-specific backends,
data aggregation from multiple downstream services per BFF, and an optimized,
purpose-built API for each client.

## What's included

- Domain models (`Product`, `CartItem`, `User`, `Order`, `SupplierRecord`)
- Downstream service interfaces + in-memory implementations standing in for
  real microservices
- `ClientBff<T>` contract + `MobileBff` / `DesktopBff` implementations
- `App.java` demonstrating both BFFs against the same user
- JUnit 5 tests for both BFFs and the demo entry point
- Class diagram (`etc/backends-for-frontends.png` + `.urm.puml`), referenced
  from the README
- README following the repo's standard pattern-page template

## Verification

- `mvn -pl backends-for-frontends -am clean test` — all tests passing
- `mvn checkstyle:check@validate -pl backends-for-frontends -am` — 0 violations
- `mvn spotless:check -pl backends-for-frontends -B` — clean
- Coverage: 98% instructions, 100% methods, 100% classes (JaCoCo)
- Full reactor build (`mvn clean install`, module excluded from unrelated flakes
  below) — green

## Notes for reviewers

- Two pre-existing, unrelated test failures were encountered while running the
  full reactor build and confirmed **not caused by this change** by reproducing
  them in isolation on a clean checkout:
  - `rate-limiting-pattern`'s `TokenBucketRateLimiterTest.shouldRefillTokensAfterTime`
    is timing-flaky.
  - `commander`'s test suite can hit `OutOfMemoryError: unable to create native
    thread` locally depending on OS thread limits, unrelated to this module.
- While verifying Checkstyle compliance, I found that `WhitespaceAround`'s
  `allowEmptyConstructors`/`allowEmptyTypes` properties are unset (default
  `false`) in the current Checkstyle config, which conflicts with Spotless/
  Google Java Format's formatting of empty `{}` bodies. This affects any record
  with a fully empty body — I counted at least 16 other existing modules with
  the same latent construct, which only avoid failing because Checkstyle isn't
  bound to the `install` lifecycle. I worked around it locally in this module's
  6 records by giving each a minimal documented compact constructor rather than
  a bare empty body, but didn't touch the shared Checkstyle config or any other
  module — happy to open a separate follow-up issue for that if it's useful.


backends-for-frontends/src/main/java/com/iluwatar/bff/model/Order.java
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * A past order returned by the order service API.
+ *
+ * @param id order identifier
+ * @param productName name of the ordered product
+ * @param status current fulfillment status, e.g. "DELIVERED", "IN_TRANSIT"
+ */
+public record Order(String id, String productName, String status) {
+  /** No additional validation required. */
+  public Order {
```
HattoriHenzo: You don't need a default constructor here. The JVM will define a default canonical constructor. The same rule applies to all the models.

backends-for-frontends/pom.xml
```diff
@@ -0,0 +1,67 @@
...
+
+  <dependencies>
+    <dependency>
+      <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter-engine</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-api</artifactId>
```
HattoriHenzo, line 48: No logging instruction in called in the code. You can remove this dependency if not using it.

HattoriHenzo (commented):
My global review or this PR:

- Add a Web Front End BFF
- In the unit tests add InMemory*.java unit tests

backends-for-frontends/pom.xml
AnveshSrivastava: Removed the compact constructors from all 6 records, good catch. They originally existed to satisfy checkstyle:check@validate (which flags empty {} bodies under WhitespaceAround), but I confirmed CI (mvnw clean verify) doesn't actually run Checkstyle, and the same pattern already exists unaddressed in converter/data-transfer-object, so I dropped the workaround to match the rest of the repo.

Also added InMemory* unit tests for all four service implementations (100% coverage on those classes now, closing the gap Codecov flagged), and confirmed slf4j-api is a genuinely required direct dependency — App.java uses it and it's not provided transitively.

On the Web BFF suggestion — issue #300's acceptance criteria only called for two client-specific backends, which Mobile/Desktop satisfy. Happy to add a Web BFF as a follow-up PR, or in this one if you'd prefer.

HattoriHenzo: Thanks for the feedback.

HattoriHenzo (commented):
A part from the build that fails, it LGTM.

AnveshSrivastava:
> A part from the build that fails, it LGTM.

iluwatar:
@all-contributors please add @AnveshSrivastava for code

PR #3584, opened 2026-08-24
build: upgrade setup-java to v6

brunoborges:
# Pull Request Template

## What does this PR do?

Upgrades all active Maven workflow references from `actions/setup-java@v4` to `actions/setup-java@v6`.

The workflow diffs were reviewed and all active `setup-java` references were checked. No runtime tests were needed for this workflow-only change.


brunoborges:
my apologies for this PR. It was only supposed to hit a set of repos with a specific dependabot configuration that would prevent getting the alert of this new release

PR #3589, opened 2026-08-27
feat: transactional outbox pattern

### LINKED ISSUE #3531: Add Outbox Pattern for Event Publishing
hey,
I've been looking through the patterns catalog and noticed you don't have an implementation for the Outbox pattern yet. I think it would be a really valuable addition because microservices and event driven architectures are very common.

the problem:
Your service needs to save something to its database and publish an event about it. But doing these two things separately means things can break. The database update might succeed, but the message broker might be down, creating zombie messages and all also causing inconsistencies.
The Outbox pattern is the fix for this. Instead of trying to do both things at once, you save the event alongside the business data in the same database transaction. Then a background process reads those events from the database and reliably pushes them out.

My plan for the implementation:
I was thinking of building a clean Spring Boot + Kafka example that shows this in action, following the project's guidelines.

What I'd include:
A new Maven module something like transactionaloutbox
The core logic: a service that saves both the business entity and the outbox record in one transaction, plus a scheduled publisher that processes pending events
A simple demo app that shows the whole flow, from creating an order to the event being consumed
Documentation: a proper README following the project's template and a UML diagram
Tests: unit tests for the core logic, integration tests with a real database and Kafka 
Docker Compose 
Why I think this is valuable

It's a real-world pattern that developers actually need
It fills a gap in the current catalog
It's a great learning resource for anyone building distributed systems
It touches on modern Java ecosystem tools like Spring, JPA, and Kafka

I've been referencing a few existing implementations:

I've been looking at the outbox-example repo (https://github.com/microservices-patterns/outbox-example) and springboot-kafka-transactionaloutbox (https://github.com/MertYilmazda/springboot-kafka-transactional-outbox) to make sure I get the details right.

I'd love to get some feedback on this idea. Does this sound like a good fit for the project? If so, I'm happy to start working on it and follow the contribution guide to get a clean PR ready.

Thanks for your time 🙌

Mukul-Howale:
Adds the **Transactional Outbox Pattern** (`transactional-outbox`) module to the pattern catalog using Spring Boot.
- **Problem**: Resolves the dual-write problem in microservices where writing to a database and publishing to a message broker in separate steps causes data inconsistencies or lost messages.
- **Solution**: Persists business entities (`Order`) and outbound events (`OutboxEvent`) into the database within a single `@Transactional` boundary. A background `@Scheduled` service (`OutboxPublisher`) periodically polls pending events and dispatches them to the message broker.
- **Key Components**:
  - `Order` & `OutboxEvent`: Spring Data JPA entities.
  - `OrderService`: `@Transactional` service performing atomic dual-writes.
  - `OutboxPublisher`: Background worker polling `PENDING` outbox events and updating status to `PROCESSED`.
  - `README.md`: Complete documentation including architecture flow, Mermaid class diagram, and programmatic examples.
  - Unit & Integration tests using `@SpringBootTest` and H2 in-memory database.

Fixes : [3531](https://github.com/iluwatar/java-design-patterns/issues/3531)


Mukul-Howale:
@iluwatar Please review

iluwatar (approved):

iluwatar:
@all-contributors please add @Mukul-Howale for code

Issue #3577, opened 2026-08-17
[Feature Request]: Add Scatter-Gather Design Pattern


priyanshuvishwakarma273403:
### Name of the Design Pattern
Scatter-Gather Pattern

### Category
Concurrency / Integration Patterns

### Intent
The Scatter-Gather design pattern is a messaging and integration pattern used to parallelize task execution. It broadcasts (scatters) a single request to multiple recipients or worker services concurrently, and then aggregates (gathers) their responses back into a single response. This significantly reduces overall response time when fetching data from multiple independent sources.

### Detailed Explanation
In enterprise architectures, a client request often requires gathering information from multiple downstream services or running multiple independent calculations. Running these calls sequentially can lead to high latency.

The Scatter-Gather pattern addresses this by:
1. **Scatter**: Spawning multiple parallel threads or asynchronous tasks (workers) to call each downstream service or execute task partitions.
2. **Gather**: Waiting for all tasks to finish (or timing out if some are slow), collecting the results, and passing them to an aggregator.
3. **Aggregate**: Merging individual responses into a unified result format returned to the client.

#### Real-World Example
Consider a travel booking aggregator (like Kayak or Expedia). When a user searches for a flight, the portal scatters the query to multiple airlines' API servers simultaneously. It gathers the prices, sorts them, and displays the complete list to the user.

### Conceptual Architecture
Here is a conceptual design for the Scatter-Gather pattern:

```mermaid
classDiagram
    direction TB
    class TaskSupplier {
        <<interface>>
        +execute() String
    }
    class AirlineService {
        -String airlineName
        -double basePrice
        +execute() String
    }
    class ScatterGatherExecutor {
        -ExecutorService executor
        +scatterGather(List~TaskSupplier~ tasks, long timeoutMs) List~String~
    }
    class Aggregator {
        +aggregate(List~String~ results) String
    }
    
    TaskSupplier <|.. AirlineService
    ScatterGatherExecutor --> TaskSupplier : scatters tasks
    ScatterGatherExecutor .. Aggregator : passes results to

bonigo1998:
Hi, I'd like to work on this issue.  
I'm planning to implement the Scatter-Gather pattern using a TaskSupplier interface, example AirlineService implementations, a ScatterGatherExecutor backed by ExecutorService, and an Aggregator to combine the results.  
[...]

PR #3590, opened 2026-08-27
feat: add scatter-gather design pattern #3577

bonigo1998:
# Pull Request Template


scatter-gather/src/main/java/com/iluwatar/scattergather/App.java
```diff
@@ -0,0 +1,24 @@
...
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
```
iluwatar, line 24: App.java seems to be empty. Usually we have here the main entry point of the program. You should supply commentary about the pattern and its implementation as well.

iluwatar: See also that it has certain level of logging so that execution can traced.

iluwatar (changes requested):

PR #3598, opened 2026-09-03
feat: add Timeout pattern (#2845)

### LINKED ISSUE #2845: Timeout pattern
**Description:**
The Timeout design pattern is crucial in a microservices architecture to handle latency and prevent cascading failures. By setting a timeout, we can avoid long waits for responses from dependent services, improving overall system resilience and performance. This pattern helps in defining a period within which a service call must complete, otherwise, it gets aborted and a fallback mechanism is triggered. This implementation is particularly important for maintaining the robustness of the system under heavy load or when some services are down or slow to respond.

**Main Elements of the Pattern:**
- **Timeout Duration:** Defines the maximum time a service call is allowed to take.
- **Fallback Mechanism:** Provides a default response or alternative workflow if a service call times out.
- **Configuration:** Ability to configure different timeouts for different services based on their response times.
- **Logging and Monitoring:** Capture timeout events for monitoring and debugging purposes.
- **Circuit Breaker Integration:** Work in conjunction with the Circuit Breaker pattern to prevent repeated calls to failing services.

**References:**
1. [Microservices Patterns](https://microservices.io/patterns/reliability/timeout.html)
2. [Resilience4j - TimeLimiter](https://resilience4j.readme.io/docs/timelimiter)
3. [Project Contribution Guidelines](https://github.com/iluwatar/java-design-patterns/wiki)
4. [Timeout pattern](https://www.vinsguru.com/timeout-pattern/)

**Acceptance Criteria:**
- Implement a configurable timeout for service calls.
- Provide a fallback mechanism for timed-out calls.
- Ensure the timeout duration can be configured per service.
- Log timeout events for further analysis and monitoring.
- Demonstrate the pattern with a practical example within the existing microservices.
- Ensure the implementation follows the project's contribution guidelines and coding standards.

ylcn91:
## What does this PR do?

Adds the **Timeout** pattern as a new `timeout` module.

- **Problem**: without a time limit a single slow dependency can hold threads and user requests hostage until the whole system stalls.
- **Solution**: every downstream call runs under a per-service `TimeoutPolicy`. When the limit is exceeded the call is cancelled with an interrupt, the event is logged and counted, and a fallback answer is returned.
- **Key components**:
  - `TimeoutPolicy` (record) and `TimeoutRegistry`: per-service configurable limits with a default.
  - `TimeoutExecutor`: enforces the limit (`Future.get(timeout)`), cancels the overrunning call, records the event in `TimeoutMetrics`, invokes the fallback. Service failures are surfaced as `ServiceCallException`, not as timeouts.
  - `ProductCatalogService` (fast) and `RecommendationService` (slow, interruptible): simulated dependencies.
  - `App`: catalog answers within its 500 ms limit; recommendations exceed their 100 ms limit, get cancelled and replaced by popular items; timeout counters are printed. Logging traces every step.
  - `README.md`: intent, real-world example, sequence diagram, code walkthrough, applicability, trade-offs, related patterns (including how this differs from the existing `fallback` module, where the time limit is only one of several triggers). PlantUML class diagram under `etc/`.
- **Tests**: 13 JUnit 5 tests covering result within limit, fallback plus cancellation on overrun (interruption verified), failure propagation, per-service limits and counters, registry and policy validation, plus `AppTest`.
- Module registered in the parent `pom.xml`. `./mvnw clean verify -pl timeout` passes locally on JDK 21 and inside an `eclipse-temurin:21` container.

Fixes #2845


ylcn91:
Note on the automated review comments: `LOGGER` is the Lombok logger field name configured for this repository in `lombok.config` (`lombok.log.fieldName = LOGGER`), the same name every other module uses, so the code compiles as is. Local `./mvnw clean verify -pl timeout` passes on JDK 21, also inside an eclipse-temurin:21 container.

ylcn91:
Follow-up on the Codecov note: added a test for an interrupted caller in `TimeoutExecutorTest` and a new `RecommendationServiceTest` (normal reply and interrupted computation). 16 tests, no uncovered lines outside `App`. Tests only, no production change. `./mvnw clean verify -pl timeout` passes locally on JDK 21 and in an eclipse-temurin:21 container.

ylcn91:
Coverage follow-up: extracted the two demo calls in `App` into package-private helpers (`loadProducts`, `loadRecommendations`) so both the normal and the fallback paths are exercised by `AppTest`, and added `shouldBeInstantiable`. Demo output unchanged. JaCoCo now reports 100% instruction, branch and line coverage. Verified locally on JDK 21 and in an eclipse-temurin:21 container; the packaged jar runs end to end.

PR #3600, opened 2026-09-03
feat: add Service-Oriented Architecture pattern (#2937)

ylcn91:
## What does this PR do?

Adds the **Service-Oriented Architecture (SOA)** pattern as a new `service-oriented-architecture` module, implemented in plain Java without frameworks so the architectural mechanism stays visible.

- **Key components**:
  - `Service`: the contract every service exposes (`name()` and `handle(ServiceRequest)`); services are stateless.
  - `ServiceRequest` / `ServiceResponse`: coarse-grained, self-describing messages that form the interoperability boundary (flat payload, uniform response envelope).
  - `ServiceRegistry`: discovery by name on a `ConcurrentHashMap`, rejects duplicate registrations.
  - `ServiceBus`: routes messages to the registered service, applies cross-cutting concerns (tracing, timing) in one place and translates provider failures into error responses, so consumers only ever talk to the bus.
  - `CustomerService`, `InventoryService`, `PaymentService`: reusable atomic services.
  - `OrderService`: composite service that orchestrates the three atomic services through the bus to place an order.
  - `App`: bootstraps registry and bus, registers the services, places an order that succeeds, one that is rejected for stock, and addresses an unregistered service. Logging traces every message.
  - `README.md`: intent, real-world example, diagram, code walkthrough, applicability, real-world uses (ESBs, SOAP/WSDL), trade-offs, related patterns including how SOA differs from microservices. PlantUML class diagram under `etc/`.
- **Tests**: 24 JUnit 5 tests covering registry, bus routing and error translation, each atomic service, and order orchestration (success, unknown customer, out of stock, payment declined), plus `AppTest`.
- Module registered in the parent `pom.xml`. `./mvnw clean verify -pl service-oriented-architecture` passes locally on JDK 21 and inside an `eclipse-temurin:21` container.

Fixes #2937

PR #3599, opened 2026-09-03
feat: add Microservices Load Shedding pattern (#3229)

ylcn91:
## What does this PR do?

Adds the **Microservices Load Shedding** pattern as a new `microservices-load-shedding` module.

- **Problem**: under a traffic spike or a slow dependency, requests pile up, latency grows for everyone and the service eventually falls over.
- **Solution**: a `LoadShedder` performs capacity-based, priority-aware admission control at the service entry point. Excess requests are rejected immediately (fail fast) instead of being queued; low-priority work is shed first and a reserve of capacity is kept for critical requests.
- **Key components**:
  - `Priority`, `Request`, `Response`: the request model with CRITICAL / NORMAL / LOW priorities.
  - `LoadShedder`: hard capacity, low-priority limit and critical reserve; lock-free compare-and-set admission so capacity is never exceeded; per-priority shed counters.
  - `LoadShedException`: the fast rejection ("503, retry later").
  - `ShedGuardedService` and `RequestHandler`: wrap the business logic and always release the capacity slot, even when the handler throws.
  - `App`: light load (everything admitted), payment provider slows down and orders pile up (LOW and NORMAL probes are shed, CRITICAL checkout still admitted), provider recovers (admission resumes), then a metrics summary. Logging traces every decision.
  - `README.md`: intent, real-world example, flowchart, code walkthrough, applicability, trade-offs, and how load shedding differs from rate limiting, throttling, backpressure and queue-based load leveling. PlantUML class diagram under `etc/`.
- **Tests**: 13 JUnit 5 tests covering admission below capacity, priority-ordered shedding, the critical reserve, release, metrics, a 50-thread concurrency test proving capacity is never exceeded, slot release on handler failure, plus `AppTest`.
- Module registered in the parent `pom.xml`. `./mvnw clean verify -pl microservices-load-shedding` passes locally on JDK 21 and inside an `eclipse-temurin:21` container.

Fixes #3229

PR #3597, opened 2026-09-03
feat: add Microservices Bulkhead pattern (#3228)

ylcn91:
## What does this PR do?

Adds the **Microservices Bulkhead** pattern as a new `microservices-bulkhead` module.

- **Problem**: when every downstream call shares one thread pool, a single slow or hanging dependency (here a payment provider) consumes all threads and healthy dependencies (inventory) start failing although nothing is wrong with them.
- **Solution**: each downstream dependency gets its own `Bulkhead`, a dedicated fixed-size thread pool with a bounded queue. When the compartment is full the call fails fast with `BulkheadFullException` instead of blocking or borrowing threads from other compartments.
- **Key components**:
  - `Bulkhead`: named, bounded `ThreadPoolExecutor` (threads + queue), fail-fast rejection, metrics (active, queued, rejected), `AutoCloseable`.
  - `BulkheadFullException`: unchecked exception carrying the compartment name.
  - `RemoteService`, `PaymentService` (slow), `InventoryService` (healthy): simulated downstream dependencies.
  - `App`: runs the same load first through one shared pool (inventory call gets rejected) and then through dedicated bulkheads (payment overflow rejected fast, inventory keeps answering), with log output tracing every step.
  - `README.md`: intent, real-world example, sequence diagram, code walkthrough, applicability, trade-offs, related patterns. PlantUML class diagram under `etc/`.
- **Tests**: 12 JUnit 5 tests (latch-based, no sleeps) covering execution, rejection when threads and queue are full, capacity release, failure propagation, thread naming, shutdown, configuration validation, plus `AppTest`.
- Module registered in the parent `pom.xml`. `./mvnw clean verify -pl microservices-bulkhead` passes locally on JDK 21 and inside an `eclipse-temurin:21` container.

Fixes #3228

PR #3601, opened 2026-09-03
feat: add Scatter-Gather pattern (#3577)

### LINKED ISSUE #3577: [Feature Request]: Add Scatter-Gather Design Pattern

ylcn91:
## What does this PR do?

Adds the **Scatter-Gather** pattern as a new `scatter-gather` module.

- **Problem**: a client request often needs answers from several independent providers; calling them one after another adds up their latencies, and one slow or failing provider should not block the whole answer.
- **Solution**: the same request is scattered to every provider concurrently, replies are gathered until each has arrived, failed or exceeded the timeout, and the successful replies are aggregated into one result. Partial results are first class.
- **Key components**:
  - `RateRequest`, `RateQuote`, `RateProvider`: a travel site asking several hotel rate providers for the same stay.
  - `InMemoryRateProvider` (fast), `DelayedRateProvider` (slow, interrupt-safe), `FailingRateProvider`.
  - `ScatterGather`: `scatter()` (one `CompletableFuture` per provider with `orTimeout`), `gather()` (waits for all to settle, keeps successes, logs timeouts and failures), `scatterGather()` chaining both with an `Aggregator`; `AutoCloseable`.
  - `Aggregator`: functional reduction of the gathered replies, with a `cheapestQuote()` factory.
  - `App`: four providers (two fast, one slower than the 300 ms timeout, one failing); logs the scatter, gather and aggregate phases and the best offer.
  - `README.md`: intent, real-world example, sequence diagram, code walkthrough, applicability, trade-offs, related patterns including an explicit note on how Scatter-Gather differs from the existing Fan-Out/Fan-In module. PlantUML class diagram under `etc/`.
- **Tests**: 13 JUnit 5 tests covering all providers replying, a slow provider dropped after the timeout while others are kept, a failing provider dropped, aggregation, empty result when nobody answers, executor shutdown, plus `AppTest`.
- Module registered in the parent `pom.xml`. `./mvnw clean verify -pl scatter-gather` passes locally on JDK 21 and inside an `eclipse-temurin:21` container.

Note: PR #3590 also targets this issue. This implementation was written independently with a different domain and design; it includes the populated, logging `App` entry point requested in the review of #3590 and treats partial results and timeouts as the core of the pattern. Happy to align with whichever direction the maintainers prefer.

Fixes #3577


ylcn91:
Note on the automated review comments: `LOGGER` is the Lombok logger field name configured for this repository in `lombok.config` (`lombok.log.fieldName = LOGGER`), the same name every other module uses, so the code compiles as is. `Thread.sleep(Duration)` is the Java 19+ overload, valid on the project's Java 21 baseline. Local `./mvnw clean verify -pl scatter-gather` passes on JDK 21, also inside an eclipse-temurin:21 container.

ylcn91:
Follow-up on the Codecov note: added two `ScatterGatherTest` cases for `close()` (interrupted caller keeps the interrupt flag; a task that ignores interrupts does not block shutdown). 15 tests, `ScatterGather` fully covered; the remaining uncovered lines are the implicit `App` constructor and the demo's empty-result lambda. Tests only. `./mvnw clean verify -pl scatter-gather` passes locally on JDK 21 and in an eclipse-temurin:21 container.

ylcn91:
Coverage follow-up: moved the result reporting in `App` into a package-private `reportBestOffer` helper so the empty-result branch is covered, and added `shouldBeInstantiable`. Demo output unchanged. JaCoCo now reports 100% instruction, branch and line coverage. Verified locally on JDK 21 and in an eclipse-temurin:21 container; the packaged jar runs end to end.

ylcn91:
@iluwatar following your note on #3577 that Scatter-Gather deserves its own module next to Fan-Out/Fan-In: this PR implements exactly that reading of the pattern. One request is broadcast to several providers, replies are gathered under a timeout, and the partial results are aggregated, with a populated, logging App entry point. CI is green and all review threads are resolved.

The other five pattern PRs (#3597, #3598, #3599, #3600, #3602) are in the same state and ready for review whenever you have time.

